PR #15, opened 2026-04-20
feat(runner): fresh-install gate, strict config schema, remove teardown

irichard00:
## Summary

PR 2 of the runner-install-UX redesign. Stacks on #14 (PR 1). Turns `pidash install` into a one-stop command and closes the `pidash remove` crash-loop gap.

### `pidash install` — three phases
1. Always write the unit file (idempotent).
2. If `config.toml` is missing on a TTY (and `--no-configure` isn't set), prompt for URL / token / name and chain into `pidash configure`. Non-TTY or `--no-configure` → print next-step hint and exit without enabling. Nothing auto-starts a runner that isn't configured.
3. Once config exists, enable + start. On Linux, print the `sudo loginctl enable-linger \$USER` hint so the service starts before interactive login.

### `service::install` split
- `write_unit(paths)` — writes the unit/plist (idempotent, no activation)
- `enable_and_start()` — `systemctl enable+start` on Linux; `launchctl bootstrap` on macOS with a `kickstart -k` fallback so re-running `install` doesn't fail on an already-loaded agent

### `pidash remove` becomes the true inverse
Stop + uninstall the service **before** touching local state, so systemd/launchd don't keep restarting a deregistered daemon with no credentials. Both service calls are tolerant of never-installed / not-running.

### Config schema tightened
- `[approval_policy]` and `[logging]` get `#[serde(default)]` — minimal TOMLs parse with the existing `Default` impls.
- `codex.model_default` is marked optional with default and skipped when `None` on serialize.
- Required fields (`runner.cloud_url` / `workspace.working_dir` / `codex.binary`) **still fail loudly on parse**; `__run` bubbles that up with action-guiding text.

### `__run` error messages
On missing `config.toml` or `credentials.toml`, bails with text pointing at `pidash configure` / `pidash install`. No more silent systemd crash loops with cryptic TOML errors.

## Why

Without the gate, `pidash install` eagerly `enable`s the unit even when there are no credentials → systemd/launchd crash loop. Without the remove extension, `pidash remove` left the service installed → same crash loop. Both are fixed here, with the schema sharpened along the way.

## Test plan

- [x] `cargo test` — 62+ tests, all green
- [x] New lib tests: minimal config with default sections parses; missing required field fails loudly
- [x] New integration test `tests/pidash_run_errors.rs` — `__run` error contract when config / creds are missing
- [ ] Manual: fresh install with TTY (chains configure), non-TTY (prints hint), `--no-configure` flag, `pidash remove` leaves machine clean

## Follow-ups

PR 3: per-workspace `runner.name` uniqueness with charset rules (runner + cloud + DB migration).

### DIFF
--- a/runner/src/cli/configure.rs
+++ b/runner/src/cli/configure.rs
@@ -29,9 +29,41 @@ pub struct Args {
     pub skip_doctor: bool,
 }
 
+/// Inputs for the core registration flow. `cli::install::run` also builds one
+/// of these — once via clap, once via interactive prompts.
+pub struct RegisterInputs {
+    pub url: String,
+    pub token: String,
+    pub name: Option<String>,
+    pub working_dir: Option<PathBuf>,
+    pub skip_doctor: bool,
+}
+
+impl From<Args> for RegisterInputs {
+    fn from(a: Args) -> Self {
+        Self {
+            url: a.url,
+            token: a.token,
+            name: a.name,
+            working_dir: a.working_dir,
+            skip_doctor: a.skip_doctor,
+        }
+    }
+}
+
 pub async fn run(args: Args, paths: &Paths) -> Result<()> {
-    validate_cloud_url(&args.url)?;
-    let name = args
+    execute(args.into(), paths, /* print_next_hint = */ true).await
+}
+
+/// Core registration flow — hits the cloud register endpoint, persists
+/// `config.toml` + `credentials.toml`, and optionally runs the doctor.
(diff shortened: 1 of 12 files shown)


irichard00:
Superseded by #18. Original base branch `feat/cli-restructure-pr1` was auto-deleted when #14 merged, which auto-closed this PR. Branch has been rebased onto `main` with review fixups folded in and re-opened as #18.

PR #44, opened 2026-04-23
fix(api): serve local dev via uvicorn so WebSockets work

irichard00:
## Summary
- `apps/api/bin/docker-entrypoint-api-local.sh` was launching Django via `manage.py runserver`, which is WSGI-only. That drops Channels' WebSocket route at `/ws/runner/`, so any locally-running runner gets `404 Not Found` on connect and stays stuck "offline" in the web UI.
- Swap the local entrypoint to `uvicorn pi_dash.asgi:application --reload`. `uvicorn` is already pinned in `apps/api/requirements/base.txt` and the prod entrypoint already uses it via `gunicorn -k uvicorn.workers.UvicornWorker`, so this aligns local dev with prod.
- `DJANGO_SETTINGS_MODULE=pi_dash.settings.local` is exported since `runserver`'s `--settings` flag has no uvicorn equivalent.

## Test plan
- [ ] `docker compose -f docker-compose-local.yml up --build api` comes up without errors and the api container serves on `:8000`.
- [ ] Plain HTTP still works: `curl -sI http://localhost:8000/` (or any known app route) returns 200/302.
- [ ] `./setup.sh && pnpm dev`, register + start a local runner via `pidash configure --url http://localhost:8000 --token <tok>`. UI at `localhost:3000` shows the runner as online within ~60s (runner's WS reconnect interval).
- [ ] Editing a Python file in `apps/api/` triggers uvicorn `--reload`.

### DIFF
--- a/apps/api/bin/docker-entrypoint-api-local.sh
+++ b/apps/api/bin/docker-entrypoint-api-local.sh
@@ -31,4 +31,8 @@ python manage.py create_bucket
 # Clear Cache before starting to remove stale values
 python manage.py clear_cache
 
-python manage.py runserver 0.0.0.0:8000 --settings=pi_dash.settings.local
+# `runserver` is WSGI-only, which drops Channels WebSocket routes (the runner
+# ↔ cloud link at /ws/runner/ returns 404). Use uvicorn directly so local dev
+# speaks ASGI, matching the gunicorn+UvicornWorker setup in production.
+export DJANGO_SETTINGS_MODULE=pi_dash.settings.local
+exec uvicorn pi_dash.asgi:application --host 0.0.0.0 --port 8000 --reload
--- a/apps/api/bin/docker-entrypoint-api-local.sh
+++ b/apps/api/bin/docker-entrypoint-api-local.sh
@@ -34,5 +34,5 @@ python manage.py clear_cache
 # `runserver` is WSGI-only, which drops Channels WebSocket routes (the runner
 # ↔ cloud link at /ws/runner/ returns 404). Use uvicorn directly so local dev
 # speaks ASGI, matching the gunicorn+UvicornWorker setup in production.
-export DJANGO_SETTINGS_MODULE=pi_dash.settings.local
+export DJANGO_SETTINGS_MODULE="${DJANGO_SETTINGS_MODULE:-pi_dash.settings.local}"
 exec uvicorn pi_dash.asgi:application --host 0.0.0.0 --port 8000 --reload
--- a/apps/api/requirements/base.txt
+++ b/apps/api/requirements/base.txt
@@ -32,12 +32,12 @@ jsonmodels==2.7.0
 django-storages==1.14.2
 # user management
 django-crum==0.7.9
-# web server
-uvicorn==0.29.0
-# WebSocket protocol for uvicorn / gunicorn's UvicornWorker. Without this
-# uvicorn can't upgrade WS requests and silently hands them to Django's
-# HTTP app, so `/ws/runner/` 404s through the URL router.
-websockets==12.0
+# web server. The `[standard]` extra pulls in websockets (required for
+# uvicorn / gunicorn's UvicornWorker to upgrade WS requests — without it
+# `/ws/runner/` 404s through the HTTP URL router), plus httptools, uvloop,
+# and watchfiles (so `--reload` uses the efficient watcher instead of the
+# legacy stat-polling fallback).
+uvicorn[standard]==0.29.0
 # sockets
 channels==4.1.0
 channels-redis==4.2.0
--- a/apps/api/requirements/base.txt
+++ b/apps/api/requirements/base.txt
@@ -34,6 +34,10 @@ django-storages==1.14.2
 django-crum==0.7.9
 # web server
 uvicorn==0.29.0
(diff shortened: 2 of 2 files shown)


irichard00 (commented):
## PR Review: #44 — fix(api): serve local dev via uvicorn so WebSockets work

**Verdict:** Approve with non-blocking comments. Root cause is diagnosed correctly and the fix is the right one — `runserver` is WSGI-only and silently drops Channels' `/ws/runner/` route, which is exactly why locally-registered runners sit "offline." Swapping to uvicorn (matching the `gunicorn -k uvicorn.workers.UvicornWorker` setup in prod) is the minimal, correct change.

### What the PR actually changes (vs. what the body says)

The PR body lists only the entrypoint swap, but the branch contains **two commits** and **two diff hunks**:
1. `docker-entrypoint-api-local.sh` — `runserver` → `exec uvicorn ... --reload`
2. `requirements/base.txt` — **adds `websockets==12.0`** (commit 5ee52e5)

Please update the PR body to mention the dep add. Reviewers shouldn't have to read the diff to notice a new runtime dependency. The inline comment in `base.txt` is a good explanation — hoist it into the PR summary.

### Scope note — this also fixes a latent prod bug (HIGH, worth calling out)

Stock `uvicorn` ships without a WebSocket implementation. You either install `uvicorn[standard]` (pulls `websockets`, `httptools`, `uvloop`, `watchfiles`) or install a WS library explicitly. Before this PR, `requirements/base.txt` pinned plain `uvicorn==0.29.0` with **no** WS lib, which means:

- Local dev (new uvicorn command): every `/ws/runner/` upgrade would `501 Unsupported upgrade request` (uvicorn logs `Unsupported upgrade request.`).
- **Production** (`gunicorn -k uvicorn.workers.UvicornWorker`): same thing — runners connecting to the prod URL would also fail WS handshake. The only reason this hasn't surfaced as a reported outage is presumably that nobody's run a runner against prod yet, or prod has been getting WS via a different path (nginx? terminating elsewhere?). Worth sanity-checking.

Adding `websockets==12.0` to `base.txt` (not `local.txt`) is the right call because of prod, but the PR framing treats it as a local-dev issue. It isn't — it's a base dependency gap. One-line mention in the PR body fixes the framing.

### Issues

**MEDIUM — `DJANGO_SETTINGS_MODULE` is hard-overwritten**
```bash
export DJANGO_SETTINGS_MODULE=pi_dash.settings.local
```
This clobbers whatever the container/compose env set. Gemini's bot review flagged the same thing. Safer:
```bash
export DJANGO_SETTINGS_MODULE="${DJANGO_SETTINGS_MODULE:-pi_dash.settings.local}"
```
Lets a dev override to `pi_dash.settings.local_debug` (or similar) via compose env without editing the script. Also consistent with how most 12-factor entrypoints do it.

**MEDIUM — `--reload` without `watchfiles` falls back to stat polling**
`websockets` alone gives you WS support, but uvicorn's `--reload` without `watchfiles` uses the legacy `StatReload` backend — slower, higher CPU, and unreliable on some bind-mount filesystems. Two options:
- Swap the pin to `uvicorn[standard]==0.29.0` (adds `websockets`, `httptools`, `uvloop`, `watchfiles` in one go). This is the most common pattern and would let you remove the separate `websockets==12.0` line.
- Or add `watchfiles` explicitly to `local.txt` (not `base.txt`, since prod doesn't need it).

I'd recommend `uvicorn[standard]` — it's literally what the extras are for, and it aligns dev+prod on a single deps story.

**LOW — Static files in local dev**
`runserver` auto-serves `STATIC_URL` via `django.contrib.staticfiles`. `uvicorn` running `asgi:application` does not, regardless of `DEBUG`. If `settings.local` doesn't mount WhiteNoise or similar, admin/DRF browsable API CSS/JS will 404 in the local container. Not a blocker for the WS fix, but verify during the test-plan checks — the `docker compose -f docker-compose-local.yml up --build api` smoke test should include a visit to `/admin/login/` to confirm assets still load.

**LOW — `exec` is correct, good catch**
Using `exec` ensures uvicorn gets SIGTERM directly from `docker stop` instead of the bash wrapper eating it. Nothing to change here, just noting that it was done right.

### Verification I ran locally

- Checked out `fix/local-api-asgi-websockets` at f4c5413 + 5ee52e5.
- Confirmed `pi_dash/asgi.py` is a proper `ProtocolTypeRouter` with `runner_ws_urls` — fix targets the right entrypoint.
- Confirmed prod entrypoint (`docker-entrypoint-api.sh`) already runs via `UvicornWorker`, so the deps fix retroactively helps prod too.
- Confirmed `docker-compose-local.yml` mounts `./apps/api:/code` so `--reload` will pick up edits.

### Action items before merge

1. Update PR body to mention `websockets==12.0` addition + the prod implication.
2. Guard the settings export: `export DJANGO_SETTINGS_MODULE="${DJANGO_SETTINGS_MODULE:-pi_dash.settings.local}"`.
3. Strongly recommend switching `uvicorn==0.29.0` + `websockets==12.0` → `uvicorn[standard]==0.29.0` (removes one line, adds `watchfiles` for proper `--reload`).
4. Smoke-test `/admin/` in the local container to confirm static files aren't silently broken.

None of the above block merging the WS fix itself — they're hardening. The core change is correct and should ship.

PR #49, opened 2026-04-23
feat(runner): introduce pods — workspace-shared runners with per-pod queue

irichard00:
## Summary

Reframes runners from personal machines (scoped to one `owner`) to **workspace-shared digital employees** grouped into **pods**. Issues pin to a pod; dispatch drains the pod's queue to idle runners. Fixes the stranded-`QUEUED` bug along the way — runners that complete a run now refire `drain_pod` so queued work moves forward without waiting for an external trigger.

Full design: [`.ai_design/issue_runner/design.md`](.ai_design/issue_runner/design.md).

## Why

- **Stranded QUEUED**: under the old matcher, if all of a user's runners were busy the run was left QUEUED with no background drain.
- **Permission model conflation**: `AgentRun.owner` was doing double duty as access principal and billable party. Once runners are shared, these identities diverge.
- **Revoke was unsafe**: `Runner.revoke()` only flipped a status bit. Force-closed runners left `ASSIGNED/RUNNING` rows hanging forever.
- **Zero-setup**: new workspaces should just work — no admin clicks required to make delegation run.

## Architecture

```
Workspace
├── Pod (many, workspace-scoped, soft-deleted)
│   ├── Runner (N, workspace-shared; owner is admin bond only)
│   └── AgentRun (N, FIFO per pod, drains to idle runners)
└── Issue.assigned_pod ──► Pod
```

Every workspace auto-creates a default pod named `<workspace.name>-pod`. Runner registration and issue creation both default to it. Pods can be deleted only when empty of runners, non-terminal runs, and not the workspace's last pod.

### Backend
- **New**: `Pod` model + `PodManager` (soft-delete), workspace post-save signal, `ensure_workspace_pods` management command, three `services/` modules (`permissions.py`, `validation.py`, pod-scoped matcher + `drain_pod`), pod CRUD endpoints.
- **Rewritten**: `Runner.revoke()` synchronously cancels in-flight runs + refires drain; `orchestration._dispatch_to_runner` → pod-aware `drain_pod_by_id`; `_resolve_owner` → `_resolve_fallback_creator`; `consumers._finalize_run` drains after completion.
- **Permission-model switch**: list/detail/cancel/approve endpoints filter by `AgentRun.created_by` (new NOT-NULL FK); runner endpoints switch from owner-scoped to workspace-scoped; revoke widened to owner-or-admin; approvals route to the run creator (not runner owner).
- **Schema**: two migrations — `runner/0004_add_pod_and_run_identity` and `db/0126_issue_assigned_pod`. Single schema cutover (no backfill ceremony) because pi-dash has no production data yet. Tolerates local dev rows via in-migration data pass.

### Frontend
- Types extended (`IPod`, `IPodMini`, runner/agent-run get `pod` + `pod_detail`, agent-run gets `created_by`).
- New `PodService` + `RunnerService.move` / `getDetail`.
- Runners page shows a Pods strip above the table and a Pod column in the table.
- i18n keys added (`runners.pods.*`, `runners.list.columns_pod`) in `en` only.

### Design doc
- Zero-setup fresh-install model with `<workspace_name>-pod` naming.
- Decision table reconciled (#3 soft-delete semantics, #7 fresh install, #9–#12 codex-feedback fixes, #13 every-workspace-has-a-pod invariant).

## Tests

**Backend**: 115/116 runner + orchestration + contract tests pass. The 1 failure (`test_run_config_carries_git_fields_from_issue_and_project`) is pre-existing on main — reproduced by stashing this PR's diff and running against main.

Test coverage added:
- `test_pod.py` — pod model, soft-delete visibility, conditional unique constraints (8).
- `test_workspace_signal.py` — auto-pod creation, idempotency, model save() auto-resolve for Runner/AgentRun (6).
- `test_permissions.py` — workspace role helpers (8).
- `test_validation.py` — every §6.5 rule + pod resolution priority (8).
- `test_drain_pod.py` — pod-scoped matcher, billing capture, FIFO (10).
- `test_revocation.py` — synchronous cancel + drain refire (6).
- `test_pod_views.py` — CRUD endpoints + all three soft-delete guards (14).
- `test_runs_views.py` — membership enforcement, auto-pod, body-spoofing protection (5).
- `test_issue_pod.py` — issue auto-pin + serializer validation (6).

**Frontend**: `@pi-dash/types`, `@pi-dash/services`, `@pi-dash/i18n` build clean. `web` `check:types` reports no new errors (2 pre-existing `sidebar-item.tsx` errors unrelated to this PR).

## Test plan

- [ ] Run the migrations against a fresh dev DB: `python manage.py migrate`.
- [ ] Run the full test suite: `pytest pi_dash/tests/unit/runner/ pi_dash/tests/unit/orchestration/ pi_dash/tests/contract/runner/`.
- [ ] Create a new workspace in dev → confirm `<name>-pod` auto-exists and is default.
- [ ] Register a runner → confirm it lands in the default pod (visible in the new Pod column).
- [ ] Create an issue → confirm `assigned_pod` is set to the default.
- [ ] Move issue to In Progress → confirm `AgentRun` is created with `pod` set and `created_by` populated.
- [ ] Revoke the runner while a run is in flight → confirm the run transitions to `CANCELLED` with `error="runner revoked"`.
- [ ] Try to soft-delete the workspace's only pod via `DELETE /api/runners/pods/<id>/` → confirm 409 `last_pod_in_workspace`.

## Deferred (called out in Phase 5 commit)

- Full pods CRUD admin UI (create/rename/delete forms).
- Move-runner dialog.
- Issue-detail pod picker (advanced section per §8.3).
- `zh`, `ja`, etc. locale files for the new `runners.pods.*` keys.

### DIFF
--- a/.ai_design/issue_runner/design.md
+++ b/.ai_design/issue_runner/design.md
@@ -26,6 +26,14 @@
 >   to reflect soft-delete preconditions; removed vestigial validation bullet
 >   and renamed `_resolve_owner` consistently to `_resolve_fallback_creator`;
 >   added billing-capture and user-deletion tests.
+> - Fresh-install simplification: no production data exists yet, so the
+>   three-stage migration (schema → backfill → tighten) collapses to a
+>   single initial schema migration. Decision #7 (opt-in cooperation) is
+>   removed — there's nothing to migrate from. New decision: every workspace
+>   auto-creates a default pod named `<workspace.name>-pod` on creation, so
+>   the first-run experience requires zero pod setup. Runner registration
+>   and issue creation both default to the workspace pod when unspecified.
+>   Soft-delete now guards against removing the last pod of a workspace.
 
 ## 1. Goal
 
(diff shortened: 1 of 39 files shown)


irichard00:
# Code review — PR #49 (pods + per-pod queue)

Scope: 41 files, +3300/-204. Reviewed backend, migrations, authz surface, and frontend in parallel. Architecture is sound and tests are substantial, but there are two authz regressions and a few correctness races worth addressing before merge.

## HIGH — should land before merge

### H1 · `_can_view_run` lets a runner's `owner` view runs regardless of workspace membership
`apps/api/pi_dash/runner/views/runs.py:30-36`

```python
def _can_view_run(user, run: AgentRun) -> bool:
    if run.created_by_id == user.id:
        return True
    if run.runner_id is not None and run.runner.owner_id == user.id:
        return True
    return is_workspace_admin(user, run.workspace_id)
```

Now that runners are workspace-shared, `Runner.owner` is an "admin bond" only — it no longer tracks current workspace membership. A user removed from a workspace (or whose membership was never refreshed) still satisfies `run.runner.owner_id == user.id` and sees every run landing on that runner. Same issue for `_can_cancel_run`. Gate on `is_workspace_member(user, run.workspace_id)` before the per-object checks; return 404 rather than 403 to avoid confirming existence cross-workspace. The `is_workspace_admin` branch is also untested.

### H2 · Pod-deletion `runners` guard counts REVOKED runners
`apps/api/pi_dash/runner/views/pods.py:165`

`PodSerializer.get_runner_count` (`serializers.py:60`) excludes `REVOKED`, but the deletion guard doesn't — so a pod whose runners were all later revoked is permanently undeletable. Switch to `pod.runners.exclude(status=RunnerStatus.REVOKED).exists()` and move it inside the atomic block (see H4).

### H3 · `consumers._finalize_run` drains outside `on_commit`
`apps/api/pi_dash/runner/consumers.py:457-460`

Design §6.3 and `Runner.revoke()` (`models.py:258`) both drain via `transaction.on_commit(...)`. `_finalize_run` calls `drain_pod_by_id` synchronously. Works today because there's no surrounding atomic block, but the moment someone wraps this in `atomic()` the WS `assign` fires before the terminal state is committed and a runner can receive a new assignment for the run it's finishing. Mirror the revoke pattern: \`transaction.on_commit(lambda: drain_pod_by_id(runner.pod_id))\`.

### H4 · TOCTOU on runner-move and pod-delete guards
- `views/runners.py:96-112` — pod-existence check and `runner.save()` aren't in the same `atomic()`. A concurrent pod soft-delete leaves the runner pointing at an inactive pod, and `drain_pod_by_id` silently skips it, stranding QUEUED runs.
- `views/pods.py:164-202` — the three delete guards run *before* `transaction.atomic()` at L194. Concurrent run/runner creation can bypass them. Move guards inside the atomic block and `select_for_update()` the pod row at the top.

### H5 · Cancel endpoint lacks `select_for_update`; races with revoke
`views/runs.py:128-151`

`Runner.revoke()` holds row locks inside `atomic`; `AgentRunCancelEndpoint` does a plain `.first()` + `.save()`. Concurrent revoke+cancel causes two `drain_pod` fires from separate on_commit hooks plus a `send_to_runner` \`cancel\` frame to a runner that may have already been revoked. Wrap the fetch in `select_for_update()` inside an explicit `atomic()` and re-check terminal state after acquiring the lock.

### H6 · Migration NOT-NULL tighten has no safety net for rows where `owner_id IS NULL`
`apps/api/pi_dash/runner/migrations/0004_add_pod_and_run_identity.py:92`

\`\`\`python
if run.created_by_id is None and run.owner_id is not None:
    run.created_by_id = run.owner_id
\`\`\`

The `AlterField` at L247 tightens `created_by` to NOT NULL. Any row with both fields NULL survives the backfill and the tighten raises `IntegrityError`. Fresh DBs are fine, but the PR description claims \"tolerates local dev rows.\" Add a pre-flight `RunPython` that raises if any \`AgentRun.objects.filter(created_by__isnull=True).exists()\` after the backfill.

### H7 · Frontend: `IRunner.pod` / `IAgentRun.pod` typed as non-nullable `string`
`packages/types/src/runner.ts:38, 85`

Backend FK uses `PROTECT` so non-null in the happy path, but with soft-delete on Pod and no compile-time guarantee, safer is `string | null`. The UI already defends (`r.pod_detail ? ... : \"—\"`), so aligning the type avoids surprise.

---

## MEDIUM

- **`runner/signals.py:41-56` swallows all errors** (`except Exception: logger.exception(...)`). Workspace creates, pod fails, invariant #13 is quietly violated until a user hits `RunCreationError(409, \"no_pod_available\")`. Surface a metric/alert or re-raise.
- **Pod name unbounded.** `signals.py:40` — `f\"{instance.name}-pod\"` can exceed `Pod.name` max_length 128. `DataError` is then swallowed by the bare except. Truncate: `f\"{instance.name[:123]}-pod\"`.
- **Back-compat shim in `AgentRun.save()` is dead on `update_fields` calls.** `models.py:399` mutates `self.created_by_id` but every `save(update_fields=[...])` in `consumers.py` / `views/runs.py:151` discards the change. Either add `\"created_by\"` to update_fields when populated, or remove the shim now that the migration has run.
- **`ensure_workspace_pods` management command has a read-then-write race.** `management/commands/ensure_workspace_pods.py:39-67` — missing-workspace list is built outside `atomic`, creates happen inside. Concurrent signal can cause `IntegrityError` that aborts the whole batch. Switch to `Pod.objects.get_or_create` per workspace.
- **`PodService.get_one()` breaks camelCase.** `packages/services/src/runner/pod.service.ts:37` — rename to `getOne` (or `getDetail` to match `RunnerService.getDetail`).
- **i18n structural inconsistency.** `packages/i18n/src/locales/en/core.ts:302` — `runners.list.columns_pod` should be `runners.list.columns.pod` to match sibling column keys.
- **`runners.pods.load_failed` is dead.** Key defined but nothing wires an `onError` handler to the pods SWR. Wire it or delete it.

---

## LOW

- Cross-workspace authz untested for `AgentRunDetailEndpoint` / `AgentRunCancelEndpoint` and the admin branch of `_can_view_run`.
- `_can_view_run` triggers an extra DB hit on `run.runner.owner_id`. Add `.select_related(\"runner\")` to the detail/cancel endpoints.
- `_backfill_run_identity` is N+1 on `run.runner.pod_id`. Immaterial for fresh DB.
- Composite index on `(pod, status)` for `AgentRun` covers the filter but not the `created_at` sort in `next_queued_run_for_pod`. Adding `created_at` as a third column eliminates the sort.
- Pods `<section>` in `apps/web/.../runners/page.tsx:200-218` has no `aria-labelledby`.
- New `runners.pods.*` keys only exist in `en`. Fallback is safe; track for the follow-up locales commit.

---

## Adjacent — pre-existing, not a regression

`RegistrationTokenCreateEndpoint` (`views/register.py`) mints a workspace runner token without checking workspace membership of the requester. Not touched by this PR, but PR #49 widens the blast radius because workspace-shared runners now receive workspace-wide work. Worth a follow-up to gate token minting on `is_workspace_admin`.

---

## What's solid

- FK `PROTECT` across `Runner.pod`, `AgentRun.pod`, `Issue.assigned_pod` is correctly audit-preserving; `AgentRun.owner` correctly relaxed to `SET_NULL`.
- Partial unique constraints handle soft-delete reuse correctly; `test_pod.py` coverage is tight.
- Body-spoofing protection on run creation (`views/runs.py:72-94`) is clean — `created_by` comes from the authenticated user via `validate_run_creation`, explicitly tested.
- `_can_manage_runner` and all pod/runner PATCH handlers correctly delegate to workspace-scoped permission helpers.
- Serializer `read_only_fields` correctly lock `workspace`, `created_by`, `runner_count`, etc.

Tests are comprehensive on the happy paths (115/116, the one failure reproduces on main). Gaps are in the concurrency paths and the admin branch of `_can_view_run`.

PR #63, opened 2026-04-26
feat(runner): auto-install `pd` shortcut during configure

irichard00:
## Summary
- During the initial register flow in `pidash configure`, drop a `pd` symlink (Unix) / file copy (Windows fallback) next to the running `pidash` binary so users can type `pd <args>` instead of `pidash <args>`.
- Skip silently when `pd` is already taken — at the target path *or* anywhere else on `PATH` — so we never shadow an existing binary (PagerDuty CLI, Pure Data, user shell aliases, etc.).
- New `runner/src/util/pd_symlink.rs` is best-effort: every failure mode (read-only fs, permission denied, Windows without symlink privileges) collapses into a `PdAction` enum the caller logs without erroring out the configure flow. Idempotent on re-run.

## Test plan
- [x] `cargo check` clean
- [x] `cargo clippy --all-targets -- -D warnings` clean
- [x] `cargo test --lib` — 75 passed (4 new tests in `util::pd_symlink::tests`)
- [ ] Manual: run `pidash configure --url ... --token ...` on a fresh dev machine where `pd` is not on `PATH` — verify shortcut is created and `pd status` works.
- [ ] Manual: re-run `pidash configure ...` and verify the second run is silent (no duplicate "Installed `pd` shortcut" line).
- [ ] Manual: install PagerDuty CLI first, then run `pidash configure` — verify the skip note prints and the existing `pd` is untouched.

### DIFF
--- a/runner/src/cli/configure.rs
+++ b/runner/src/cli/configure.rs
@@ -472,6 +472,8 @@ pub async fn execute(inputs: RegisterInputs, paths: &Paths) -> Result<()> {
     };
     crate::config::file::write_credentials(paths, &creds)?;
 
+    report_pd_shortcut(crate::util::pd_symlink::try_install());
+
     if !inputs.skip_doctor {
         run_doctor_with_auth_gate(paths, agent_kind).await?;
     }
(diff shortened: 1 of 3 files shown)


irichard00:
Closing — not needed right now. Branch `feat/runner-pd-shortcut` left intact for future reference; reopen this PR or branch from it if we revisit the `pd` / `pi` shortcut idea.

PR #92, opened 2026-04-30
refactor(runner-tui): bus + view-stack + per-tab state + declarative keymap

irichard00:
## Summary

Replaces the runner TUI's monolithic event loop (~520-line `handle_event`, flat `AppState`, per-iteration `terminal.draw()`) with the layered architecture described in `.ai_design/tui_refactor/design.md`. Fixes three concrete bugs that all stem from the same missing structure:

1. **Digit / `h` / `l` keys swallowed by the global tab switcher while typing into form fields.** Each form field is now a `TextArea` widget; while one is focused the active keymap context is `[TextInput]` only, so digit/letter keys cannot resolve to `GoToTab` / `NextTab`.
2. **List selection resets every 500ms refresh tick.** Each tab owns a `SelectableList<Id>`; on every IPC update the dispatcher reconciles against the new item list — selection follows the original element by identity across re-orderings, additions, deletions.
3. **Focus indicator only visible on the Runners tab.** Each tab now has a `Pane` enum, the single source of truth for visual focus; the border-style helper is applied across every multi-pane tab.

## Architecture

Modeled on codex-rs (`tui::Tui` terminal owner, `AppEvent` bus, `FrameRequester` coalesced redraws) and claudy (focus as derived state, declarative keybindings with scoped contexts, leaf-first input routing, identity-stable selection via `validatedFocusedValue`). See `.ai_design/tui_refactor/design.md` for the full architecture lift and `.ai_design/tui_refactor/tasks.md` for the phase-by-phase plan.

### New modules
- `tui_runtime::Tui` — sole owner of stdout, raw mode, bracketed paste, focus events. Panic hook + `Drop` restore the terminal.
- `tui_runtime::TuiEventStream` — physical-event source: filters `KeyEventKind::Release`, drops Mouse, maps FocusGained → Draw, exposes bracketed-paste passthrough.
- `tui_runtime::FrameRequester` + `FrameScheduler` — coalesced redraws with 120fps clamp; replaces per-loop `terminal.draw()`.
- `event::AppEvent` + `event_sender::AppEventSender` — single application bus; widgets hold a clone of the sender, never reach into `App`.
- `view::View` + `view::tab::Tab` — modal stack contract + base-tab contract, with `Renderable` shared as the leaf primitive.
- `widgets::ScrollState` (ported from codex), `widgets::SelectableList` (claudy `validatedFocusedValue` semantics), `widgets::TextArea` (single-line, `KeyEventKind::Release` filtered at the boundary).
- `input::keymap::{Action, Context, KeymapRegistry, resolve}` — declarative `(KeyEvent, Context) → Action` registry; the active contexts list is computed per dispatch with most-specific-first ordering. `Context::TextInput` is intentionally empty so leaf textareas always win.
- `input::paste_burst` — pure state machine (vendored from codex) for terminals without bracketed-paste support.
- `views/modals/{confirm,help,add_runner,remove_runner}` — each modal is a `View` impl pushed onto the stack and auto-popped on completion.

### Per-tab state
`AppData` carries cross-cutting application data (IPC snapshots, loaded config, banners). Per-tab structs (`GeneralTab`, `RunnerStatusTab`, `RunsTab`, `ApprovalsTab`) own their pane focus, list cursors, TextAreas, and edit buffers — the old flat `AppState` is gone. The 500ms ticker is the sole cadence owner; IPC polls run on spawned tasks per `Tick` gated by per-concern `*_in_flight` flags.

## Test plan
- [x] `cargo build` — release build clean
- [x] `cargo test` — 138 lib tests pass (23 new unit tests covering ScrollState, SelectableList, TextArea, paste_burst, resolver), 38 integration tests pass
- [x] `cargo clippy -- -D warnings` — clean
- [ ] Manual smoke (per `.ai_design/tui_refactor/tasks.md` checklist):
  - [ ] Type \`1234567890\` into Cloud URL while register form is focused — all 10 chars inserted; no tab switches
  - [ ] Type \`hl\` into hostname — both characters inserted, not tab navigation
  - [ ] Tab between cards on every tab — visual focus indicator follows
  - [ ] Open Approvals with ≥ 20 items, scroll past visible window, wait one tick — selection retained
  - [ ] Paste a multi-line URL into Cloud URL — inserted as one chunk
  - [ ] Open any modal → \`Esc\` closes; global keys (\`q\`, digit keys) inert while open
  - [ ] Resize during refresh — no panic, no corrupted glyphs


irichard00 (commented):
## Code Review — refactor/runner-tui-architecture

Re-reviewed after merging latest `main` into the branch. Verdict: **APPROVE with one blocker (cargo fmt) and a few cleanup items.**

The architecture is sound. All three claimed bug fixes are structurally correct: the `Context::TextInput`-only resolution prevents digit/`h`/`l` escape into tab switches, `SelectableList<Id>::reconcile()` preserves selection identity across IPC ticks, and per-tab `Pane` enums make focus a derived property rendered consistently across all multi-pane tabs. Bus-based dispatch + view-stack + leaf-first input routing match the codex-rs/claudy reference patterns the design doc cites.

`cargo check` clean. `cargo clippy -D warnings` clean. `cargo test` 205 passed / 0 failed.

### Blocker

**`cargo fmt --check` fails.** Diffs in `runner/src/agent/mod.rs` (lines 382, 438, 454, …) and `runner/src/tui/widgets/textarea.rs:201`. Run `cargo fmt` before merge.

### Re-verification of the prior gemini-code-assist review

| # | Finding | Status on current HEAD |
|---|---------|------------------------|
| 1 | `paste_burst.rs:81-83` infinite loop on non-ASCII re-feed | **Code path exists, but module is unwired.** `PasteBurst::on_plain_char` has no caller in the source tree (only `pub mod` declaration in `input/mod.rs` + its own `#[cfg(test)]` block). The loop cannot occur at runtime. The non-ASCII branch is still incorrect (pushes `c` into the buffer before returning `Flush(s)`) — fix when wiring it up, or drop the module. |
| 2 | `add_runner.rs` projects never written back, picker unreachable | **Resolved.** The current modal (329 lines) has no `projects` field, no `ensure_projects_loaded`, no `Picker`. Module docstring is explicit: "The cloud assigns workspace + project from the one-time token, so no client-side project picker." |
| 3 | `design.md:513` lists `ipc_listener.rs` contradicting §5.7 | **Confirmed, self-acknowledged at line 996.** Doc-only; no runtime impact. Worth a follow-up cleanup commit but not merge-blocking. |

### New findings

**HIGH — `mask_token` panics on multi-byte input** — `runner/src/tui/views/modals/add_runner.rs:326`
```rust
format!("{}…{}", &raw[..2], &raw[raw.len() - 2..])
```
Byte-index slicing. If a token contains any multi-byte UTF-8 char in the first two or last two bytes, this panics with a byte-boundary split and crashes the TUI. Real enrollment tokens are ASCII, but a stray paste from a clipboard with smart quotes/em-dash would crash. Use `chars()`/`char_indices()` or sanitize input upfront.

**MEDIUM — `expect()` in render/key paths** — `runner/src/tui/views/config.rs:656`, `runner/src/tui/views/runner_status.rs:278`
Both call `FIELDS.iter().position(|f| f.id == id).expect(...)` from paths reachable during render or key handling. If a `FieldId` variant is added without a corresponding `FIELDS` entry, the TUI panics in the render hot path. The panic hook restores the terminal, but it's a maintenance trap for a future contributor. Use `unwrap_or_else(|| { tracing::warn!(...); 0 })` or a static-asserted `EnumMap`.

**MEDIUM — Spawn outlives its view (double-pop hazard)** — `runner/src/tui/views/modals/add_runner.rs:123`, `runner/src/tui/views/modals/confirm.rs:158`
The submit task captures a cloned `AppEventSender` and posts `AppEvent::PopView` on completion. If the user presses Esc to cancel before the task finishes, the view is already popped via `Cancellation`. When the task then completes, its `PopView` pops whatever view is on top of the stack now — typically a different modal or the base view. Consider tagging events with the originating `ViewId` and ignoring `PopView` for stale ids, or storing a cancellation flag the spawn can observe.

**MEDIUM — Unbounded `mpsc` for the AppEvent bus** — `runner/src/tui/app.rs:1123`
`unbounded_channel()` is fine in the happy path (the `select!` is biased to drain `rx` first), but a hung dispatch under heavy IPC traffic accumulates `Tick` and IPC events without bound. Bounded `mpsc::channel(256)` would give a natural ceiling and surface back-pressure as a slow-handler signal rather than memory growth.

**LOW — `KeyEventKind` filter inconsistency** — `picker.rs:139` checks `Press` only; `approvals.rs:152` / `runs.rs:94` / `general.rs:252` / `runner_status.rs:188` accept `Press | Repeat`. Held-key navigation works everywhere except the picker. The stream boundary at `event_stream.rs:63` already filters `Release`, so the per-handler re-checks are redundant defense-in-depth — make them consistent.

**LOW — `TabKind::from_idx(...).unwrap()`** — `app.rs:957,961`. The unwrap is provably safe (`% 4`), but an `expect("tab idx is always 0..3")` would document why.

### Architecture spot-checks

- **Background task lifetime**: All `tokio::spawn` callsites capture only `tx: AppEventSender` and `ipc: TuiIpc` (both `Clone` + owned). No task holds a borrow into `App` or `AppData`. Sender drop on shutdown converts in-flight events into silent drops — correct per design.
- **Render purity**: `render_overlay` in `app.rs:329` takes `&mut Frame` (necessary for `StatefulWidget`) but no concrete `View::render_overlay` implementation observed mutates view state. Convention holds.
- **No `Mutex` in TUI hot path**: All TUI mutation sits on the single `select!` thread. No deadlock surface.

### API contract drift from main merge — clear

Spot-checked the slug-or-id rewrite added on main against runner callers. `runner/views/projects.py` (the runner-facing project list) is outside the slug rewriter; `cli/issue.rs`, `cli/resolve.rs`, etc. continue to source `project_id` as UUIDs from API responses and pass them verbatim. `Project.resolve()` accepts both forms, so existing UUID traffic is unchanged. No drift.

### Action items before merge

1. **`cargo fmt`** (blocker)
2. Fix `mask_token` byte slicing or guard against multi-byte input
3. Replace `expect()` in render/key paths with safer fallbacks
4. (Follow-up) Decide: wire `PasteBurst` into `TextArea` or remove the module
5. (Follow-up) Patch `design.md:513` module-layout diagram

The pytest CI failure on this PR appears to predate the just-merged main (which fixed the `project-slug-or-id` contract tests). Re-running CI after `cargo fmt` should clear it.

PR #102, opened 2026-05-03
feat(prompting): fork agent flow on coding vs non-coding task type

irichard00:
## Summary

- The agent prompt previously framed every issue as a coding task and unconditionally ran `git fetch origin` + branch setup in Step 1.7 and `commit` / `push` / PR-open in Step 2. Non-coding issues (e.g. *"run `pwd` and post a comment"*) would still trigger git operations and could hang on credential prompts or unrelated remote state.
- Adds a **Task type** classification (`code_change` | `noncode`) to the Step 0.5 analysis. The agent self-classifies based on whether its proposed approach edits files in the repo.
- Gates the git sync, branch creation, commit/push, and PR-opening sub-steps on `code_change`. `noncode` runs go straight from workpad setup → executing the task → final comment → state move.
- Defaults to `code_change` when uncertain — the heavier path is the safer default if the agent misclassifies.

## Files changed

- `01_intro.md` — drops "coding agent" framing; introduces the two paths.
- `07_analyze_and_scope.md` — adds the **Task type** bullet to the Step 0.5 analysis and updates the Proceed gate to mention noncode skipping.
- `08_workpad_setup.md` — gates step 1.7 (git fetch / branch setup) on `task_type == code_change`.
- `09_implementation.md` — gates step 5 (commit/push) and step 6 (open PR) on `task_type == code_change`.
- `12_workpad_template.md` — adds **Task type** to the workpad `### Analysis` section so the classification is auditable across continuation runs.

## Test plan

- [ ] Trigger a non-coding test issue (e.g. *"run `pwd` and post a comment via pidash CLI"*) and verify the runner does **not** invoke `git fetch origin` or any branching/commit/push commands.
- [ ] Trigger a coding issue (small file edit) and verify the agent still runs the full git sync → branch → commit → push → PR flow.
- [ ] Confirm the workpad's `### Analysis` section now includes a `**Task type**` line for both runs.
- [ ] Spot-check one ambiguous issue to confirm the agent defaults to `code_change` rather than dropping into `noncode` and skipping the PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #104, opened 2026-05-04
feat(cli): make pidash CLI read URL/workspace/token from config file

irichard00:
## Summary

Fixes the `PIDASH_API_URL is not set` failure that's been silently killing every agent run that touches the `pidash` CLI.

The pidash CLI's CRUD subcommands (`pidash issue/comment/state/workspace`) demand three env vars:
- `PIDASH_API_URL`
- `PIDASH_WORKSPACE_SLUG`
- `PIDASH_TOKEN`

**Nothing in the runner spawn path sets them.** The agent's prompt template tells it to use `pidash` for all cloud interaction. The first `pidash` call exits 1 with `{\"error\":\"PIDASH_API_URL is not set\"}`. The agent has no way to read the issue or post results, codex emits an empty `turn/completed`, and the run fails.

## Approach

The CLI binary is the same binary as the daemon. The daemon already has `~/.config/pidash/config.toml` with `cloud_url` (under `[daemon]`) and `workspace_slug` (under the primary `[[runner]]`). The CLI can read them too — no env-var plumbing needed.

- New `[cli]` table in the config schema with one field, `token`. For v1 the token is populated manually by the operator (a future cloud endpoint can mint per-run scoped tokens from the runner's refresh credentials).
- `CliEnv::resolve(paths)` reads URL/workspace/token from the config file, falling back to the corresponding env var per field. Error messages name the missing config key so partial configs are diagnosable.
- The 4 CRUD command sites switch from `from_env()` to `resolve(&paths)`.
- `from_env()` retained for callers without a `Paths` handle.

## What this is NOT

- **No cloud-side token mint endpoint.** The token is still operator-populated. The natural follow-up is an endpoint that mints a scoped, per-run token using the runner's refresh credentials so the operator never has to copy a token by hand.
- **No per-run env vars** like `PIDASH_ISSUE_IDENTIFIER`. That's a separate small change to the runner spawn path; not needed to fix the immediate bug.
- **No \"one workspace per daemon\" validation.** The CLI reads workspace from the *primary* runner; multi-workspace daemons would currently see only the first workspace's slug. Tightening to a single workspace per daemon is a config-validator change worth doing but separate.

## Test plan

- [x] `cargo test --lib` — 171 passing (+1 for the new resolver path).
- [ ] After merge: populate `[cli] token = \"…\"` in `~/.config/pidash/config.toml` on the dev host, restart pidash, trigger a run that calls `pidash issue get`, confirm the call succeeds.

## Stack

Branched off `design/ticking-drop-resume`. Once that lands, this PR rebases onto wherever the chain ends up.

### DIFF
--- a/runner/src/api_client.rs
+++ b/runner/src/api_client.rs
@@ -60,10 +60,23 @@ impl std::fmt::Display for CliError {
 
 impl std::error::Error for CliError {}
 
-/// Runtime configuration sourced from the environment.
+/// Runtime configuration the CRUD subcommands need to talk to the cloud.
 ///
-/// The agent receives these from the runner-injected env (PR 2). For
-/// out-of-band operator use, the same vars can be exported in the shell.
+/// **Resolution order** (see `CliEnv::resolve`):
+///   1. The runner config file at `~/.config/pidash/config.toml`.
+///      `api_url` ← `[daemon].cloud_url`,
+///      `workspace_slug` ← primary `[[runner]].workspace_slug`,
+///      `token` ← `[cli].token`.
+///   2. If the file is missing, OR any required field is absent from
+///      it, fall through to the corresponding env var
+///      (`PIDASH_API_URL` / `PIDASH_WORKSPACE_SLUG` / `PIDASH_TOKEN`).
+///
+/// This means: an enrolled host doesn't need any env vars — the CLI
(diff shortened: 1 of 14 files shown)


irichard00:
Superseded by #109 — content landed on main via cherry-pick after the #102/#103 stack rerouted.

PR #111, opened 2026-05-05
feat(cli): add `pidash issue create`

irichard00:
## Summary
- Adds `pidash issue create --project <SLUG|UUID> --title "..." [--description] [--priority] [--state]` to create a project-scoped work item from the CLI.
- `--project` is required — the CLI is machine-global and intentionally has no default project, so callers must always name one.
- `--state` reuses the existing `resolve_state_name(client, project_ref, ...)` helper, so name-or-UUID works the same as on `issue patch`.
- Adds a short note in `resolve.rs` explaining why there's no `resolve_project` helper: project-scoped REST routes accept either a UUID or the workspace-scoped slug, so the CLI passes `--project` straight through.

## Stacked on #112
This PR is **stacked on top of #112** (`feat(api): accept project slug or UUID across all project-scoped routes`).

- Base: `api/project-slug-or-id` (will auto-retarget to `main` once #112 merges).
- The diff shown here is only the Rust CLI delta; the Python API changes belong to #112.
- `--project <SLUG>` requires #112 to be merged. `--project <UUID>` works against `main` today.

**Merge order: #112 first, then this.**

## Test plan
- [ ] `cargo build -p pi-dash-runner`
- [ ] Manual: `pidash issue create --project <UUID> --title "smoke"` returns the created work-item JSON on stdout, exit 0
- [ ] Manual: `--title ""` exits with `EXIT_INVALID` and an error on stderr
- [ ] Manual: `--state "In Progress"` resolves to a UUID via `resolve_state_name` and is sent in the body
- [ ] After #112 merges: re-run with `--project ENG` and confirm slug routing works

PR #112, opened 2026-05-05
feat(api): accept project slug or UUID across all project-scoped routes

irichard00:
## Summary
Project-scoped REST URLs previously required a UUID in the URL path. The URL converter `<uuid:project_id>` rejected slugs at routing time, before any view ran, forcing every client to know the project's UUID upfront. This widens **177 routes** (51 in `/api/v1/`, 126 in `/app/v1/`, plus the project detail PK) to accept either form transparently.

## Mechanism
- **Single source of truth:** `Project.resolve(workspace_slug, value)` detects UUIDs via `uuid.UUID()`, otherwise looks up by `identifier__iexact` scoped to the workspace.
- **Pre-permission hook:** `BaseAPIView.initial` and `BaseViewSet.initial` (both `app/` and `api/` trees) rewrite `kwargs["project_id"]` (or `kwargs["pk"]` for the project detail route) to the canonical UUID **before** DRF runs `check_permissions`, so existing permission classes and view code see UUIDs unchanged.
- **OpenAPI:** postprocess hook annotates the now-string-typed `project_id` and project `pk` parameters with a description noting both forms are accepted, plus UUID and slug examples.

## Tests
- 12 routing tests in `apps/api/pi_dash/tests/api/test_project_identifier_routing.py` covering one route per URL file (UUID/slug equivalence) plus 404, mixed-case, cross-workspace isolation, and an `/api/` smoke for the parallel app tree.
- Full suite shows no new regressions vs `main`.

## Why
Enables CLI / automation to pass human-readable slugs (`ENG`) instead of UUIDs. First consumer is #111 (`pidash issue create --project ENG`).

## Test plan
- [ ] CI: `pytest apps/api/pi_dash/tests/api/test_project_identifier_routing.py`
- [ ] CI: full Django test suite (no regressions)
- [ ] Manual: `GET /api/v1/workspaces/<ws>/projects/<UUID>/work-items/` works (existing behavior)
- [ ] Manual: `GET /api/v1/workspaces/<ws>/projects/ENG/work-items/` works (new behavior)
- [ ] Manual: `GET /api/v1/workspaces/<ws>/projects/eng/work-items/` works (case-insensitive)
- [ ] Manual: cross-workspace slug lookup returns 404
- [ ] Manual: OpenAPI schema renders the dual-form description on `project_id`


irichard00 (commented):
## Review summary

Reviewed via three specialized agents (python, security, database) plus a Codex feedback validation pass. Two CRITICAL and three MEDIUM/LOW findings were applied in commit `c782a41`; remaining items are out-of-scope follow-ups.

### Applied (commit `c782a41`)

| Severity | Finding | Fix |
|---|---|---|
| CRITICAL | **Anonymous slug-existence oracle.** `_rewrite_project_kwarg` ran before DRF's `perform_authentication`, so an anonymous request could distinguish "slug missing" (404) from "auth missing" (401) without credentials. | Guard the rewrite on `view.request.user.is_authenticated` in both `api/views/base.py` and `app/views/base.py`. The lazy auth check still triggers DRF's authenticators; `super().initial()` -> `check_permissions` then 401s the unauthenticated case uniformly. |
| CRITICAL (perf) | **`identifier__iexact` cannot use the existing btree.** Every slug-form request would seq-scan `projects` on the hot path of 177 routes. | `Project.save()` already normalizes to upper, so the resolver now does `identifier=str(value).strip().upper()` — equality match uses the existing composite uniqueness index `project_unique_identifier_workspace_when_deleted_at_null`. |
| MEDIUM | **Http404 message echoed in response.** DRF's `exception_handler` converts `Http404(*args)` → `NotFound(*args)`, propagating raw input + workspace slug into the response body. | Replaced with generic `"Project not found"`. |
| Test gap | No coverage for the unauthenticated-slug response code (would have caught the oracle). | Added `test_anonymous_slug_returns_401_not_404` asserting both real and bogus slugs yield 401 for anonymous callers. |

### Confirmed clean

- Cross-workspace slug isolation: `Project.resolve` correctly scopes via `workspace__slug=workspace_slug`.
- UUID-shaped slug attack: `identifier` is `max_length=12`, can't hold a 36-char UUID.
- UUID fast-path correctly skips the DB roundtrip (intentional, documented).
- Composite index already exists via `UniqueConstraint(fields=["identifier", "workspace"], condition=deleted_at__isnull=True)`.
- `MultipleObjectsReturned` blocked by the same unique constraint plus `Project.save()` upper-casing.

### Out-of-scope follow-ups

- Stash resolved `Project` instance on `request` to avoid the second fetch in downstream `get_object`.
- Extract the duplicated `_rewrite_project_kwarg` into `pi_dash.utils.views`.
- `dispatch()` exception arm in three base classes returns `exc` instead of `response` — pre-existing, not exercised by this PR's new path.

### CI note

`Test API` is red, but it's been red for every run since the workflow was introduced 9 days ago (0/48 green). Repo-wide environment issue (missing `APP_BASE_URL`/`WEB_URL`, fixture isolation), not a regression from this PR.

PR #113, opened 2026-05-06
feat(cli,prompt): pidash issue list + expose create/list to agent

irichard00:
## Summary
Two coupled changes:

### 1. `pidash issue list`
Adds `pidash issue list --project <SLUG|UUID> [--cursor <c>] [--per-page <N>] [--order-by <field>]` to the runner CLI. Hits the existing `GET /api/v1/workspaces/<slug>/projects/<project>/work-items/` endpoint (the same URL as `issue create`, just with GET) and emits the server's full pagination envelope:

\`\`\`json
{ "count": 42, "next_cursor": "...", "prev_cursor": null, "results": [ ... ] }
\`\`\`

Callers walk pages by passing the prior page's `next_cursor` back in. Slug or UUID both work via the `Project.resolve` rewrite hook landed in #112.

Implementation notes:
- 12-line inline percent-encoder for query values (unreserved set per RFC 3986 §2.3). Preferred over adding `urlencoding` / `percent-encoding` as a direct dep for one call site.
- 5 unit tests cover unreserved/reserved/multibyte and query-string join behavior.
- `--project` is validated for empty-string client-side, mirroring `issue create`.

### 2. Prompt update — expose `issue create` and `issue list` to the agent
The default agent prompt fragment `03_pidash_cli.md` currently only documents `issue get` and `issue patch`. Adds entries for `issue list` and `issue create` so the agent can:
- **`issue list`**: scan related/duplicate issues in the same project before creating new ones, or before drawing context-poor conclusions.
- **`issue create`**: file follow-up issues for **discovered scope that does not belong on the current issue**. Sub-steps of the current run still belong in the workpad — that guardrail is preserved.

A regression test in `test_fragments.py` asserts both markers stay present so neither command is silently dropped from a future fragment edit.

## Operator step
Existing instances need `./manage.py reseed_default_template` to pick up the new fragment text. New instances seed correctly on first run.

## Test plan
- [x] `cargo test --lib` (181 passed, 0 failed; 5 new tests for the percent-encoder and query-string builder)
- [x] `cargo check` clean on `runner/`
- [x] Fragment markers verified present via `python3 -c "..."` smoke
- [ ] Manual: `pidash issue list --project <UUID>` returns paginated JSON envelope, exit 0
- [ ] Manual: `pidash issue list --project ENG --per-page 10` accepts slug, returns ≤10 results
- [ ] Manual: walk two pages with `--cursor` from `next_cursor`
- [ ] Manual: `--project ""` exits with `EXIT_INVALID`
- [ ] Manual: re-seed default template on a dev instance and confirm a turn-start prompt mentions both new commands


irichard00 (commented):
## Review summary

Decision: **APPROVE with comments** — small, focused PR; CLI matches existing `cmd_create`/`cmd_patch` patterns; new percent-encoder is RFC 3986 §2.3-correct and has solid unit coverage; fragment regression markers are the right shape.

Validation:
- `cargo check` (runner) ✅
- `cargo clippy --no-deps` ✅ for `cli/issue.rs` (9 pre-existing lib warnings live in `daemon/state.rs` + `daemon/supervisor.rs`, not introduced here)
- `cargo test --lib` ✅ — 181 passed, including the 5 new tests
- pytest `test_fragments.py` not runnable in my env (missing `celery`); verified the asserted markers (`pidash issue create --project`, `pidash issue list --project`) are present in the on-disk assembled fragment body via direct file concat (same shape as `assemble()`)

## Findings

### MEDIUM — `--project` flows into the URL path raw

`runner/src/cli/issue.rs:197-200`:
```rust
let path = format!(
    "workspaces/{}/projects/{}/work-items/{query}",
    client.env.workspace_slug, project_ref
);
```
`project_ref` is `args.project.as_str()` and `CliEnv::v1` doesn't URL-encode the path either. This isn't introduced by this PR (`cmd_create` does the same since #112), but this PR exposes both `list` and `create` to the agent prompt, so it widens the agent attack surface. An agent-supplied `--project ../../../some/other/endpoint` would compose into the URL; `url::Url` will normalize it before send and could reach unintended `/api/v1/...` endpoints under the same `X-Api-Key`.

Suggested fix (apply to both `cmd_list` and `cmd_create`): reject any `project` value that isn't a UUID or a slug matching `^[A-Za-z0-9_-]+$` before interpolating, or run the path segment through the new `percent_encode_value` helper.

### MEDIUM — trim mismatch between validation and value sent

`runner/src/cli/issue.rs:180-183` rejects `args.project` when `trim().is_empty()`, then line 183 sends the **untrimmed** value into the URL. `--project "  ENG  "` passes validation, gets URL-encoded as `%20%20ENG%20%20`, server 404s. Cheap fix: `let project_ref = args.project.trim();`. Same pattern in `cmd_create`.

### LOW

- L1: `args.cursor.as_ref()` then `c.clone()` (`issue.rs:186-188`) — cosmetic; `args.cursor.clone()` flows the same way.
- L2: `format!("%{:02X}", b)` allocates per byte (`issue.rs:231`). Not a hot path; leave as-is.
- L3: `per_page: u32` is unbounded client-side. Server enforces a cap; not worth duplicating.

## Notes (non-issues)

- Percent-encoder is exactly the RFC 3986 §2.3 unreserved set — slightly more conservative than necessary for query values, but conservative is the right default for a one-shot encoder. Tests cover unreserved/query-separators/multibyte.
- URL shape, error mapping, JSON output, and exit-code conventions all match the existing `cmd_get`/`cmd_create`/`cmd_patch` shape.
- Fragment update keeps the "use the workpad for sub-steps; create only for follow-up scope" guardrail explicit, which matters now that the agent can call `issue create`.
- Operator step (`./manage.py reseed_default_template`) is correctly documented in the PR description.

## Recommended next steps

1. Apply the M2 trim fix (1 line in `cmd_list` + matching change in `cmd_create`).
2. Decide on M1 scope — either accept the pre-existing pattern, or factor a `validate_project_ref` helper and wire it into all three sites (`cmd_list`, `cmd_create`, plus the state-resolution path in `cmd_patch`). The latter is the safer call now that the agent prompt names both commands.
3. Run the four unchecked manual items from the PR description before merge.

PR #123, opened 2026-05-12
PDASHOSS01-3 add .pidash/ dir into the workdir when adding runners

irichard00:
## Summary

- When a new runner is enrolled (`pidash connect` or the TUI add-runner modal), write `<working_dir>/.pidash/context.md` describing which cloud / workspace / project this checkout is now linked to and which local runner owns the link.
- The marker file lives next to the user's repo (alongside `.git`) so a workdir is self-describing about its Pi Dash project, independent of the runner config file which lives under `$XDG_CONFIG_HOME/pidash/`.
- New `runner/src/workspace/context.rs` module with the writer + 4 unit tests. Both enrollment paths in `cli/connect.rs` (`run` and `enroll_additional_runner`) call the writer after the local config has been persisted.
- Marker-write failures are logged (`tracing::warn` + a `Warning:` stderr line) but do **not** unwind enrollment, because the cloud-side registration and local config have already landed and the enrollment token is single-use.

### `.pidash/context.md` content

```
# Pi Dash workspace context

This directory is linked to a Pi Dash project. ...

- Cloud: https://pidash.example.com
- Workspace: acme
- Project: PDASHOSS01
- Runner: laptop-1 (12345678-1234-5678-1234-567812345678)
- Linked at: 2026-05-11T...
```

### Scope notes

- Revoke / remove flow is intentionally **not** changed in this PR — the issue asks for adding the marker on enroll; deletion semantics (audit-trail vs. clean-removal) can be a follow-up.
- The "re-enroll after revive" branch (which returns early after rotating credentials.toml only) is intentionally not touched: the workdir is unchanged so the existing marker is still correct.

## Test plan

- [x] `cargo check` clean
- [x] `cargo clippy --lib` clean
- [x] `cargo test --lib` — 216 passed
- [x] `cargo test --tests` — all integration suites pass
- [x] New unit tests cover: file placed at `<wd>/.pidash/context.md`, missing workdir is created, body contains all identifying fields, rewrite overwrites previous contents.
- [ ] End-to-end manual: run `pidash connect --url ... --token ... --working-dir /path/to/repo` against a real cloud and confirm `/path/to/repo/.pidash/context.md` appears.


irichard00:
Superseded — `.pidash/context.md` is already written on main via `runner/src/cli/context.rs` (the `pidash context init` subcommand + `write_context_for_project(...)` helper that landed with PR #128).

Closing in favor of a smaller follow-up that wires the existing helper into the new `pidash runner add` / `pidash connect` enroll path so the marker is dropped automatically, instead of reintroducing a parallel `workspace/context.rs` module with a different schema.

PR #167, opened 2026-05-26
feat(web): gate password-change UI on is_self_managed

irichard00:
## Summary

Hides the user "Security" tab (change-password form) on hosted/managed deployments where account credentials are owned by an upstream identity provider, not by pi-dash.

- `apps/web/core/components/settings/profile/sidebar/item-categories.tsx` — filters the `"security"` entry out of the profile-settings sidebar when `config.is_self_managed === false`.
- `apps/web/app/(all)/settings/profile/[profileTabId]/page.tsx` — redirects `/settings/profile/security` to `/settings/profile/general` under the same condition, so a bookmarked URL or stale in-app link can't surface the change-password form on hosted instances.

Self-hosted users continue to see the tab and use the form as before. Hosted deployments that set `IS_SELF_MANAGED = False` server-side stop rendering it automatically — no override needed in the cloud overlay.

This is the same seam already used by [#165](https://github.com/The-AI-Republic/pi-dash/pull/165) (hiding the in-app "Deactivate account" button on hosted deployments). The motivation here is identical: hosted deployments delegate credential lifecycle to the upstream IdP (AI Republic SSO in the cloud case), so an in-app password-change form is misleading at best and broken at worst.

Tracked in private-pi-dash issue PRIVATEPI1-2.

## Test plan

- [ ] Self-managed instance (`IS_SELF_MANAGED=True` / default): "Security" tab still appears in the user settings sidebar; `/settings/profile/security` still loads the change-password form.
- [ ] Hosted instance (`IS_SELF_MANAGED=False`, e.g. private-pi-dash cloud overlay): "Security" tab no longer appears in the sidebar; direct nav to `/settings/profile/security` redirects to `/settings/profile/general`.
- [ ] `pnpm check:types` clean in `apps/web`.


irichard00:
Closing because this hosted-deployment behavior belongs in private-pi-dash's cloud overlay, not OSS pi-dash.

PR #178, opened 2026-05-27
fix(runner): make start() wait for daemon to actually exec

irichard00:
## Summary
Follow-up to #177. Same root cause family (async launchctl), different code path.

- `launchctl bootstrap` and `launchctl kickstart -k` both return as soon as launchd accepts the request, **before** the new daemon process has been exec'd. The `kickstart -k` path is the practical pain point: when the previous daemon is slow to die (active agent run in flight, slow network call, post-wake socket drain), the full SIGTERM → exit → respawn cycle can take 10+ s, but the launchctl invocation returns in milliseconds.
- Callers like `restart_and_verify` then started their 5 s IPC-verification clock from a \"daemon is up\" assumption that wasn't yet true. Users saw a scary `Daemon did not answer IPC within 5s after restart` error even though the restart actually succeeded a few seconds later. Confirmed in the wild after #177 shipped.
- `start()` now snapshots the pre-action PID, runs bootstrap (or kickstart fallback) as before, then polls `launchctl list` until it reports a fresh PID — meaning a positive PID different from the pre-action one. Now the IPC clock starts when the daemon is genuinely running.
- 30 s wait budget — covers launchd's worst-case `TimeoutTerminateSec` (20 s default) + reap + exec, with headroom. A healthy restart returns in well under a second.
- On timeout we surface a clean error but do **not** signal the daemon: at that point launchd has accepted the request, and a genuinely-mid-init daemon is its problem to report (next `LastExitStatus`), not ours to race-kill.

## Who this affects
Anyone on macOS running `pidash restart` while the daemon is slow to exit. Real cases from the field:
- Restarting during an active agent run (daemon drains in-flight runs on shutdown).
- Flaky or slow internet, or a cloud incident (every user enters the same network-retry backoff and every restart is slow).
- macOS sleep/wake cycle leaves pending socket timeouts to drain.

## Scope
macOS-only — `runner/src/service/launchd.rs`. Linux/`systemctl` is already synchronous for both `start` and `restart`; Windows/`schtasks /Run` is too.

## Test plan
- [x] `cargo test -p pidash --lib service::launchd::` — 9/9 pass (no new unit tests; the change is in subprocess-driven async code, covered by the existing `parse_loaded_pid` parser tests plus the manual scenarios below)
- [x] `cargo check -p pidash` — clean
- [ ] On macOS, repro path: induce a slow-shutdown daemon (e.g. active agent run, or unreachable cloud + restart) and confirm `pidash restart` returns a clean `daemon restarted (...)` success line instead of the false `runner failed to come up` error.
- [ ] On macOS, happy path: idle daemon, `pidash restart` completes well under 2 s with no regression.
- [ ] On macOS, fresh install: `pidash install` (which calls `enable_and_start` → `start()`) succeeds — `pre_pid = None` branch.
- [ ] On Linux: smoke-test `pidash restart` — cfg-gated, but worth verifying no surprises.

### DIFF
--- a/runner/src/service/launchd.rs
+++ b/runner/src/service/launchd.rs
@@ -16,6 +16,15 @@ const STOP_GRACE: Duration = Duration::from_secs(10);
 const STOP_POLL_INTERVAL: Duration = Duration::from_millis(150);
 const POST_KILL_GRACE: Duration = Duration::from_millis(500);
 
+/// Upper bound on how long `start()` will wait for `launchctl bootstrap`
+/// (or the `kickstart -k` fallback) to actually produce a running daemon
+/// before giving up. 30s covers launchd's worst-case shutdown cycle for
+/// the kickstart path: SIGTERM → up to TimeoutTerminateSec (20s default
+/// on macOS) → SIGKILL → reap → exec, plus headroom. A healthy restart
+/// returns in well under a second.
+const START_GRACE: Duration = Duration::from_secs(30);
+const START_POLL_INTERVAL: Duration = Duration::from_millis(150);
+
 /// Write the LaunchAgent plist. Does NOT bootstrap (load) it; that's deferred
 /// to `enable_and_start` so `pidash install` can gate activation on
 /// `pidash configure` completing first.
@@ -117,6 +126,20 @@ pub async fn uninstall(_: &Paths) -> Result<()> {
 /// failure (plist missing, malformed, permission denied, …) is surfaced
 /// with launchctl's own stderr text so the operator can diagnose it.
 ///
+/// Once the launchctl call succeeds, this waits until `launchctl list`
+/// reports a fresh PID before returning. Both `bootstrap` and
+/// `kickstart -k` are asynchronous — they return as soon as launchd
+/// accepts the request, *before* the new process has been exec'd. The
+/// `kickstart -k` case is especially bad: launchctl returns immediately,
+/// but the actual SIGTERM → exit → respawn cycle can take 10+ s when
+/// the previous daemon is wedged (active agent run, slow network call,
+/// post-wake socket drain). Callers like `restart_and_verify` then start
+/// their IPC-verification clock from a "daemon is up" assumption that
+/// wasn't true yet — surfacing a confusing
+/// `Daemon did not answer IPC within 5s` even though the daemon comes up
+/// fine a few seconds later. Waiting here moves that clock-start to the
+/// moment the daemon is actually running.
+///
 /// Why the `launchctl list` check matters: `launchctl bootstrap` reports
 /// the "you raced an in-flight teardown" case as the opaque
 ///   "Bootstrap failed: 5: Input/output error"
@@ -138,6 +161,10 @@ pub async fn start() -> Result<()> {
     let target = format!("{domain}/{LABEL}");
     let plist = plist_path()?;
 
+    // Snapshot the pre-action PID so `wait_for_running` can tell a fresh
+    // daemon from the one we're about to kick. None = no daemon before.
+    let pre_pid = probe_loaded_pid().await.filter(|p| *p > 0);
+
     let bootstrap = Command::new("launchctl")
         .arg("bootstrap")
         .arg(&domain)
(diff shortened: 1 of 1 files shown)


irichard00 (commented):
## Code Review — APPROVE

Solid, narrowly-scoped follow-up to #177 that closes a real race the team has
already seen in the field: `launchctl bootstrap` / `kickstart -k` return as
soon as launchd accepts the request, so `restart_and_verify`'s 5 s IPC clock
was starting against a daemon that hadn't exec'd yet — surfacing a misleading
"did not answer IPC within 5s" while the restart actually succeeded a few
seconds later.

The fix snapshots the pre-action PID, runs bootstrap (or the kickstart
fallback), then polls `launchctl list` until it reports a positive PID
different from the snapshot. The 30 s budget is correctly sized against
launchd's worst-case `TimeoutTerminateSec` (20 s default) + reap + exec
with headroom.

### What I checked

- **Correctness**: `is_fresh_pid` predicate is right across all four reachable
  states — install (`pre = None`, any positive PID accepted), restart
  (`pre = Some(N)`, only a different positive PID accepted), label-listed-no-PID
  (`current = Some(0)`, keep polling), label-gone (`current = None`, keep
  polling). The four unit tests cover all four. The "another caller restarts
  the daemon between our snapshot and our bootstrap" concurrency case is
  addressed in the rustdoc and the predicate still holds (any later PID we
  see differs from the snapshot).
- **Cancellation/safety**: `tokio::time::sleep` in the poll loop is
  cancel-safe; `wait_for_running` doesn't signal the daemon on timeout
  (rationale documented inline — SIGKILL here would race a mid-init
  daemon, and launchd's `LastExitStatus` is the right surface).
- **Pattern compliance**: mirrors `stop()` / `wait_for_unload`'s shape
  (deadline + poll + escalation-or-bail), uses the existing `probe_loaded_pid`
  parser, and the new `START_GRACE` / `START_POLL_INTERVAL` constants live
  alongside `STOP_GRACE` / `STOP_POLL_INTERVAL`. Comments explain *why*, not
  what — consistent with the rest of this file.
- **Security**: no user-supplied input enters launchctl args; target is
  built from `LABEL` (const) and numeric UID. No new attack surface.
- **Scope**: macOS-only via `#[cfg(target_os = "macos")]`. Linux/systemd and
  Windows/schtasks are already synchronous for start/restart, as the PR
  description states.

### Validation

| Check | Result |
|---|---|
| `cargo check -p pidash` | Pass |
| `cargo clippy --all-targets` | Pass (clean) |
| `cargo test --lib` (runner crate) | 255/255 |
| `cargo test --lib service::launchd::` | Module is `#[cfg(target_os = "macos")]`; only compiled on macOS. PR description reports 9/9 pass there. |

Toolchain: rustc 1.93.1 (pinned). The earlier gemini-code-assist let-chains
flag is moot — let-chains stabilized in 1.88, and the second commit
(`b8ba8f7f`) already refactored that predicate into the `matches!`-based
`is_fresh_pid` helper.

### Minor (non-blocking)

1. `let label_loaded = pre_pid.is_some() || probe_loaded_pid().await.is_some();`
   in `runner/src/service/launchd.rs` — `pre_pid` is filtered to `> 0`, so
   `pre_pid.is_some()` already implies "the daemon was loaded a moment ago"
   and this short-circuit just saves one `launchctl list` invocation. Worth
   a one-line comment so a future reader doesn't wonder whether it's racing
   the new daemon launchd may be exec'ing.
2. No integration test for `wait_for_running` itself (you note this in the
   PR description — `parse_loaded_pid` and the new `is_fresh_pid` tests do
   cover the pure logic, and the subprocess seam is hard to test without
   refactoring for a stub). Not blocking; flagging for the next time this
   function gets touched.
3. The PID-reuse caveat in `is_fresh_pid`'s doc is the right YAGNI call,
   given macOS's PID-walk behavior and the 30 s safety-net timeout. No
   action needed — just calling it out so the next reader doesn't try to
   "fix" it.

Nice work — the comments around the *why* (kickstart async semantics, why
we don't SIGKILL on timeout, why the fallback isn't unconditional) are
exactly the kind of context this file benefits from.

PR #202, opened 2026-06-03
PDASHOSS01-5 add support for cursor cli

irichard00:
## What

Adds **Cursor Agent** (`cursor-agent`) as a third agent backend for the runner, alongside Codex and Claude Code. A runner can now be configured to drive Cursor's CLI coding agent for assigned issue runs and chat.

Resolves PDASHOSS01-5 ("add support for cursor cli").

## Approach

The runner already abstracts agent CLIs behind `AgentBridge`/`AgentCursor` (enum dispatch over Codex + Claude Code). Cursor's `cursor-agent --print --output-format stream-json` emits a stream-json format very close to Claude's, so the new `cursor_agent` module mirrors `claude_code`:

- **`schema.rs`** — stream-json event shapes: `system/init`, `user`, `assistant`, `tool_call` (started/completed), `result`, with an `Unknown` catch-all for forward-compat.
- **`process.rs`** — subprocess wrapper. No stdin: the prompt rides in argv.
- **`bridge.rs`** — `spawn`/`warm`/`run`/`next_events`/`interrupt`/`shutdown` + translation of events into the agent-agnostic `BridgeEvent` shape.

### Key design decision: one-shot, lazy spawn

Unlike Claude (long-lived process fed turns over stdin), `cursor-agent` print mode takes the prompt as a **positional argument** and exits after emitting `result`. So the subprocess is **one-shot** and spawned **lazily at `run`** (the prompt isn't known at `spawn` time). To keep the supervisor flow unchanged, the exit-watch channel and stderr ring are created up front in `Bridge::spawn` and handed to the lazily-spawned child, so `process_handle()` — which the supervisor captures *before* the first `run` — observes the real child once it starts, and `bridge_has_exited()` (chat path) reads a consistent liveness signal (reset to `None` at each `run`). The reported PID is `None` until the first `run` (observability only, behind the opt-in `agent_observability_v1`, off by default).

### Permissions & auth

Runs with `--force` (allow commands unless explicitly denied), mirroring Claude Code's MVP `bypassPermissions` posture — approvals aren't wired yet (`send_approval` errors, same as Claude). Auth is out-of-band via `cursor-agent login` / `CURSOR_API_KEY`, like Codex/Claude.

## Wiring

- `AgentKind::CursorAgent` + `CursorAgentSection` (binary defaults to `cursor-agent`); per-runner `[runner.cursor_agent]` block is `#[serde(default)]` so existing `config.toml` files still parse.
- Selectable via `pidash runner add --agent cursor-agent`, the TUI agent cycle (codex → claude-code → cursor-agent), and editable cursor binary/model TUI fields.
- `pidash doctor` gains a `cursor-agent` version check + auth hint.
- 15-minute stall timeout (same as Claude — tool calls can be silent).

## Testing

- New `tests/cursor_bridge_fake.rs`: drives a fake `cursor-agent` subprocess through `init → assistant → tool_call → result` (happy path, one-shot, error→Failed, and warm-returns-resume-id-without-spawning).
- `schema.rs` unit tests for event dispatch.
- `cargo clippy --all-targets --locked -- -D warnings` clean; `cargo test --all-targets --locked` green (308 lib + integration).


irichard00:
## Review pass (automated)

Reviewed the cursor-agent integration end-to-end against the `claude_code` bridge it mirrors. Overall this is a clean, faithful mirror — dispatch (`AgentBridge`/`AgentCursor`), config (`CursorAgentSection`, serde-default), `doctor`, CLI `--agent cursor-agent` (ValueEnum → kebab), and TUI agent-cycle are all wired consistently. One concrete bug found and fixed; the two prior gemini findings reconciled below.

### Confirmed bug — fixed in 3ada1a8b
**TUI config editor never exposed the cursor-agent fields.** `FieldId::CursorBinary` / `FieldId::CursorModelDefault` were plumbed through `as_key`/`from_key`/`display_value`/`set_text_value`, but **not added to the `FIELDS` registry** (`runner/src/tui/views/config.rs`) that drives the rendered + navigable field list. Codex and Claude Code each contribute two `FieldSpec` entries; Cursor contributed zero, so the binary/model_default could not be viewed or edited from the config view and those two match arms were effectively unreachable. clippy didn't catch it because the variants are still constructed by `from_key`. Fixed by adding the two `FieldSpec` entries under a "Cursor Agent" section. clippy clean, `cursor_bridge_fake` + lib tests green.

### Reconciling the gemini review

Both gemini findings target `cursor_agent/process.rs`, and both describe code that is a **byte-for-byte mirror of the shipped `claude_code/process.rs`** (`shutdown` else-branch unbounded `exit_rx.changed().await`; `wait_task` `None` arm awaiting natural exit without `start_kill`). So neither is a regression introduced by this PR — they're shared latent patterns. My assessment:

- **shutdown "deadlock" (gemini: high) — not a live hang, but genuinely more fragile here than in Claude.** It does *not* deadlock today: the bridge passes `self.exit_rx.clone()`, and the bridge never advances its own `exit_rx` (no `changed()`/`borrow_and_update()` on it), so the clone's seen-version stays at the initial value and `changed()` returns immediately once any exit snapshot is published. Worth noting though: unlike `ClaudeProcess` (which owns the only `exit_tx`, so a dropped sender gives `changed()` an `Err` backstop), the **Cursor bridge keeps `exit_tx` alive through `Bridge::shutdown`**, removing that backstop. So the safety here rests entirely on "exit_rx is never advanced" — a property that would silently break if anyone later calls `changed()` on the bridge's copy. gemini's defensive shape (`if exit_rx.borrow().is_some() { return }` + `while exit_rx.borrow().is_none()` loop) is cheap hardening. **Recommend** applying it (ideally to both bridges for parity) rather than treating it as a blocker.

- **orphan/zombie on channel drop (gemini: medium) — low impact, mirrors Claude.** The `None` arm awaits natural exit without `start_kill`. Because cursor-agent print mode is one-shot, by the time a `CursorProcess` is dropped/replaced it has almost always already emitted `result` and exited, so `child.wait()` returns immediately. The orphan window only opens if the bridge is torn down mid-turn via `Drop` (not the async `shutdown`/`interrupt` paths). gemini's `let _ = child.start_kill();` in the `None` arm is a safe, correct robustness improvement — again best applied to both bridges for consistency.

Neither process.rs item is a merge blocker in my view; I did **not** auto-apply them to avoid diverging cursor from the sibling claude bridge unilaterally. Flagging for the author's call (a small follow-up touching both bridges would be the clean way to land them).

irichard00:
## Review-of-review pass — one correction + fix applied (`ef67d066`)

Re-reviewed the cursor-agent integration **and** the prior automated review pass. The integration itself is solid (dispatch, config serde-defaults, doctor, `--agent cursor-agent`, schema/process/bridge mirror of `claude_code`). One correction to the prior pass:

### The TUI config-editor fix (`3ada1a8b`) didn't actually surface the fields

The prior pass stated the cursor `binary`/`model_default` fields were invisible because they were missing from `FIELDS`, while "Codex and Claude each expose both." That premise is incorrect. The **only** renderer of the config view — `editable_lines` (`runner/src/tui/views/config.rs`, called once at `runner_status.rs:451`) — draws a hardcoded curated subset: `Runner` → name, `Workspace` → working_dir, `Agent` → kind, `Approval policy` → 3 toggles. It renders **none** of the six agent binary/model fields — not Codex, not Claude, not Cursor.

So adding the Cursor pair to `FIELDS` (`3ada1a8b`) made it **navigable** in the focus tree (`focus_tree` at `runner_status.rs:113` maps every `FIELDS` entry to an interactive item) but it still **did not render** — the cursor fields remained invisible/uneditable in the TUI, the same as the (pre-existing) Codex/Claude gap.

This is also a real pre-existing bug on `main`: those six fields are keyboard-navigable but un-rendered, so an operator can land on — and edit — an invisible field with no selection highlight. (`allowlist_commands`/`denylist_commands`, by contrast, are deliberately `readonly_row`s and are *not* in `FIELDS` — confirming `FIELDS` is the intended-editable set, so binary/model belong in the rendered view.)

### Fix applied (`ef67d066`)
`editable_lines` now renders `Codex` / `Claude Code` / `Cursor Agent` sections (binary + model_default each), in `FIELDS` order so the index-based selection highlight stays aligned. This completes `3ada1a8b`'s intent, makes all three agents' fields visible/editable consistently, and closes the navigable-but-invisible gap for Codex/Claude too. `cargo clippy --all-targets --locked -- -D warnings` clean; 308 lib tests + 4 `cursor_bridge_fake` tests green.

### On the two gemini `process.rs` findings
Agree with the prior pass's reconciliation: both target code that is a byte-for-byte mirror of the shipped `claude_code/process.rs` (verified — claude's `wait_task` `None` arm also awaits natural exit without `start_kill`; claude's `shutdown` has the same unbounded `changed().await`). Neither is a regression. The shutdown one doesn't hang today (bridge never advances its own `exit_rx`); the orphan-on-`Drop` one is low-impact given the one-shot lifecycle. Both are cheap hardening best landed on **both** bridges in a follow-up — non-blocking, left to author discretion.

irichard00:
## Independent re-review pass — validated, approve

Independently re-reviewed the cursor-agent integration and the prior review pass (the `ef67d066` TUI fix). No new defects; confirming the prior conclusion.

**Validated `ef67d066` (the TUI render fix) is correct and complete:**
- `FIELDS` has 12 entries; `editable_lines` now renders all of them (Runner → Workspace → Agent → Codex/Claude Code/Cursor Agent binary+model pairs → Approval bools) in `FIELDS` order, so the index-based selection highlight stays aligned with `index_of(id)`.
- Confirmed against `main`: `editable_lines` there rendered only Runner/Workspace/Agent/Approval, so the prior pass's premise is correct — `3ada1a8b` made the cursor fields focus-navigable but unrendered, and Codex/Claude had the same latent gap. Now closed for all three.
- No remaining navigable-but-unrendered field: `LogLevel` is intentionally out of `FIELDS` (General tab); `allowlist/denylist` are `readonly_row`s. Every `FIELDS` entry is now drawn.

**`process.rs` gemini findings — reconciliation confirmed:**
- Diffed `cursor_agent/process.rs` vs `claude_code/process.rs`: `wait_task` (incl. the `None` arm the orphan finding targets) is line-for-line identical, and `shutdown` differs only in `exit_rx` ownership (a consequence of the lazy one-shot spawn). Neither finding is a cursor-specific regression — both are optional, best-applied-to-both-bridges hardening, as already flagged.
- Minor record correction: describing `process.rs` as a "byte-for-byte mirror" is loose at the file level (it differs structurally — no stdin, lazy spawn, prompt-in-argv). Only the cited `wait_task`/`shutdown` regions are true mirrors. Doesn't change the non-regression conclusion.

**Verdict:** Approve. The one real defect (agent binary/model fields not rendered) is fixed and pushed; the two `process.rs` items remain optional follow-ups at author discretion.

irichard00:
## Added: Cursor in the cloud "Add runner" dropdown (commit `1fa35c4a`)

Following up on the human request to surface cursor in the cloud UI (the gap flagged in the prior review passes). Folded the web change into this PR rather than a separate one so the feature lands **atomically** — the dropdown option and the runner's acceptance of `--agent cursor-agent` ship together, leaving no window where the cloud UI could emit a command the released runner rejects.

**Change (`apps/web/core/components/runners/add-runner-modal.tsx`):**
- `AGENT_OPTIONS` → `["claude-code", "codex", "cursor-agent"]`. The value `cursor-agent` is exactly what `clap::ValueEnum` derives for `AgentKind::CursorAgent`, so the generated `pidash runner add --agent cursor-agent` is valid against this PR's CLI.
- `agentOptionLabel` gains a `"Cursor"` branch.
- Added the source-English `"Cursor"` i18n key in `packages/i18n/src/locales/en/translations.ts`, mirroring the existing explicit `"Claude Code"` / `"Codex"` entries.

The modal is a pure command generator (no backend call — `agent` flows straight to `--agent` in `runner-cli-command.tsx`), so this is a client-only completion of the integration; the in-file comment already mandated keeping this list in sync with `schema.rs:AgentKind`.

**Verification:** `tsc --noEmit` reports no errors in the changed files (the repo's pre-existing typecheck failures are unrelated — `TStateGroups`, `github-sync`, `scheduler.store`, etc.). The pre-commit oxfmt hook also reflowed a few previously non-compliant long lines in the same file; the functional change is the `AGENT_OPTIONS` + `agentOptionLabel` edit.

> Note: the two optional `process.rs` hardening items from the earlier gemini review (shutdown `changed()` guard; `start_kill()` on channel-drop) are still open and best applied to both the cursor and claude bridges — non-blocking, left to author discretion.

PR #206, opened 2026-06-03
PDASHOSS01-17 comment & run ai shouldn't pop up another dialogue for user to type

irichard00:
## What & why

Fixes **PDASHOSS01-17**. The "Comment & Run" button in the issue activity comment toolbar opened a *second* modal with its own textarea, asking the user to type again — even though they had already typed in the comment composer. Since the comment itself is the prompt for the run, that second input was redundant and confusing.

Now the button posts whatever the user has already typed in the composer and immediately triggers a `comment_and_run` agent run. No second dialog.

The plain **Comment** submit is unchanged (posts only; the run waits for the next tick).

## Changes

- **`comment-create.tsx`** — extract the comment-posting logic into a `postComment` helper (returns the created comment), add a `submitComment` callback that posts the current composer content, and let `extraToolbarActions` be a render-prop receiving `{ isEmpty, isSubmitting, submitComment }`.
- **`comment-and-run-button.tsx`** — drop the modal; on click, `submitComment()` then `triggerRun({ mode: "comment_and_run" })`. Disabled when the composer is empty; skips the run if the post fails.
- **`comment-and-run-modal.tsx`** — deleted (and removed from the barrel export).
- **`issue-activity/root.tsx`** — wire the render-prop; drop now-unused `issueServiceType` plumbing (kept on the prop type for caller compatibility).

## Validation

- `oxlint` on the changed files — 0 warnings, 0 errors.
- `apps/web` `check:types` — 0 errors in changed files (the 42 reported errors are pre-existing on `main`, verified by stashing the change).


irichard00:
## Review pass — concurrency finding addressed

I reviewed this PR and independently confirmed the race the Gemini bot flagged: `submitComment` runs imperatively, bypassing react-hook-form's `handleSubmit`, so the form's `isSubmitting` never went `true` during a post. A rapid double-click on **Comment & Run** could post the comment twice and dispatch two agent runs (duplicate work + cost).

Fixed in `2c48fbb4`:
- **`comment-create.tsx`** — added an `isPosting` flag set across `postComment`, merged with the form's `isSubmitting` into a single busy flag that now gates the editor (Enter-to-submit) and toolbar actions; `submitComment` no-ops while busy. `postComment` is memoized so `submitComment`'s deps are sound (dropped the `eslint-disable` for stale closures).
- **`issue-activity/root.tsx`** — the Comment & Run button is now disabled while `isSubmitting`.

Validation: oxlint clean on changed files; `apps/web` `check:types` shows 0 errors in changed files (42 pre-existing errors on `main`, unchanged). The four bot review threads are resolved.

The rest of the PR (removing the redundant modal, render-prop wiring, deletion of `comment-and-run-modal.tsx`) is correct and matches the issue intent. LGTM with this fix.

🤖 Review by [Claude Code](https://claude.com/claude-code)

PR #211, opened 2026-06-09
PDASHOSS01-10 add parent issue content as context for the issue when sending to agent

irichard00:
## What

When Pi Dash builds the prompt for an issue that has a **parent**, include the parent's context so the agent inherits the framing the sub-issue sits inside. Implements the rules from PDASHOSS01-10:

1. **Direct parent** — inline the parent's description content, report its **comment count** (count only, not the bodies), and give the exact ready-to-run command to read them: `pidash comment list <parent-id>`.
2. **Grandparent or higher** — instead of dumping every ancestor's content, render the **lineage tree** `current (current) → parent → grandparent → … → root`, and tell the agent to run `pidash issue get <ancestor>` / `pidash comment list <ancestor>` to fetch any ancestor's details on demand. Only the direct parent's content is inlined.

## Changes

- **`prompting/context.py`** — the `parent` context dict now also exposes `description` and `comments_count`; a new top-level `lineage` key carries the `current → … → root` chain, set **only** when the chain is deeper than two levels (a single parent needs no separate tree). The ancestor walk (`_ancestor_chain`) guards against accidental FK cycles with a visited-id set and a depth cap.
- **`prompting/fragments/01_intro.md`** — renders the parent-context block (description, comment count, comment-list command) and, when a grandparent exists, the lineage chain plus the ancestor-CLI guidance.

## Testing

`pytest pi_dash/tests/unit/prompting/` → **37 passed** (run in a `python:3.12-slim` container against Postgres 15.7, matching CI). 6 new tests cover: parent description + comment count in context; `lineage` is `None` for a single parent and populated/ordered for a grandparent; the rendered parent block (content + count + `pidash comment list`); the rendered lineage tree + ancestor CLI prompt; and no parent block for an independent issue. No regressions.

### DIFF
--- a/apps/api/pi_dash/prompting/context.py
+++ b/apps/api/pi_dash/prompting/context.py
@@ -32,6 +32,43 @@ def _issue_description_markdown(issue: Issue) -> str:
     return ""
 
 
+def _issue_identifier(issue: Issue) -> str:
+    """Workspace-scoped identifier, e.g. ``TP-12``.
+
+    Always uses the issue's *own* project identifier — a parent may live in a
+    different project than its child, so this must not assume the child's
+    project.
+    """
+    return f"{issue.project.identifier}-{issue.sequence_id}"
+
+
+def _issue_comment_count(issue: Issue) -> int:
+    """Count of comments on ``issue`` — matches what ``pidash comment list``
+    returns. Surfaced for a parent so the agent learns the discussion volume
+    without inlining the parent's comment bodies into this prompt.
+    """
+    from pi_dash.db.models.issue import IssueComment
+
+    return IssueComment.objects.filter(issue=issue).count()
+
+
+def _ancestor_chain(issue: Issue) -> list[Issue]:
+    """Return ``[issue, parent, grandparent, ... root]``.
+
+    Walks the ``parent`` self-FK upward. The FK has no DB-level acyclicity
+    guarantee, so defend against accidental cycles with a visited-id set and a
+    hard depth cap — a malformed graph must never spin the renderer.
+    """
+    chain: list[Issue] = []
+    seen: set[Any] = set()
+    current: Issue | None = issue
+    while current is not None and current.id not in seen and len(chain) < 50:
+        chain.append(current)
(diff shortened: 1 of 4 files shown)


irichard00:
## Review — approve (with minor optional notes)

Reviewed the implementation against PDASHOSS01-10. It correctly satisfies all of the issue's rules:

- **Direct parent** → inlines the parent description, reports the comment **count** only (not bodies), and gives the exact `pidash comment list <parent-id>` command.
- **Grandparent or higher** → renders the `current (current) → parent → … → root` lineage and points the agent at `pidash issue get/comment list <ancestor>`, inlining only the direct parent's content. `lineage` is correctly `None` for a single parent.

Verified beyond the happy path:
- Both `pidash issue get` and `pidash comment list` are real, documented subcommands (see `fragments/03_pidash_cli.md`), and the prompt passes the correct `PROJ-N` identifier form.
- `_issue_comment_count` uses the default (`objects`) manager, which filters soft-deleted rows — so the count matches what `pidash comment list` actually returns. ✅
- `parent` and `lineage` are unconditionally present in `build_context`, and every render path (`composer.py`, `views.py`) goes through it — so the new template vars are safe under the renderer's `StrictUndefined`. ✅
- The `_ancestor_chain` cycle guard (visited-set + depth cap) is sound.

**Non-blocking notes:**
1. The two N+1 nits from @gemini-code-assist (use `select_related('project')` in the ancestor walk; reuse `ancestors[1]` for `parent`) are valid micro-optimizations. Impact is negligible in practice — chains are 1–2 deep and hard-capped — so fine to apply or skip at the author's discretion.
2. Minor: the depth cap of 50 silently truncates a pathological deep chain, in which case `lineage[-1]` would be labeled the 'root' command but isn't the true root. Real-world hierarchies never approach this, so cosmetic only.

I did not auto-apply changes: the perf nits are optional and I can't run the prompting suite in this environment (no local Python/Postgres). The author's reported run shows **37 passed (6 new)**; the new tests are well-formed and internally consistent.

LGTM. 🤖

PR #212, opened 2026-06-09
PDASHOSS01-7 let's currently hide Stickies feature from user

irichard00:
## What

Hide the inherited **Stickies** feature (from Plane) from end users. Pi Dash is an agent-oriented platform and doesn't surface Stickies.

The sidebar nav item, the customize-navigation toggle, and the home dashboard widget render are **already** hidden on `main`. This PR closes the two remaining user-visible surfaces:

1. **`/stickies` route** — was still reachable by typing the URL directly. Removed its registration in `apps/web/app/routes/core.ts`. The page/layout/header components are intentionally left in place so the feature can be restored by simply re-adding the route.
2. **Manage Widgets modal "Your stickies" toggle** — the modal enumerates every server-provided widget key, so it rendered a stickies toggle even though the dashboard widget's component is already `null`. Added a `HIDDEN_HOME_WIDGET_KEYS` set in `home-dashboard-widgets.tsx` and applied it in `widget-list.tsx`.

No backend, store, or service code is removed — the feature remains fully restorable.

## Testing

- `pnpm turbo run check:types --filter=web` — the three changed files introduce no new type errors (the remaining failures are a pre-existing baseline, confirmed by stashing the changes and re-running: identical failures in `github-sync` / `readonly/state` / `scheduler.store`, none in the changed files).
- `oxlint --deny-warnings` on the three changed files — 0 warnings, 0 errors.

## Notes

- Sidebar and customize-navigation already exclude Stickies on `main`; no change needed there.
- The `allStickiesModal` command-palette path is only triggered by `StickyActionBar`, which is dead code (never imported), so it is not user-facing.

PDASHOSS01-7

### DIFF
--- a/apps/web/app/routes/core.ts
+++ b/apps/web/app/routes/core.ts
@@ -103,10 +103,9 @@ export const coreRoutes: RouteConfigEntry[] = [
           ),
         ]),
 
-        // Stickies
-        layout("./(all)/[workspaceSlug]/(projects)/stickies/layout.tsx", [
-          route(":workspaceSlug/stickies", "./(all)/[workspaceSlug]/(projects)/stickies/page.tsx"),
-        ]),
+        // Stickies — hidden from users (PDASHOSS01-7). The page/layout/header
+        // components are kept in place so the feature can be restored by
+        // re-registering this route; it is intentionally not reachable by URL.
 
         // Prompts (prompt templates: view for members, create/edit for workspace admins)
         layout("./(all)/[workspaceSlug]/prompts/layout.tsx", [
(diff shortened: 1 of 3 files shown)


irichard00:
Addressed the empty-state edge case flagged here in 18d48ae1c.

Confirmed real: `isAnyWidgetEnabled` in the home store is `Object.values(widgetsMap).some(w => w.is_enabled)`, so an enabled-but-hidden `my_stickies` (null component) would keep the dashboard branch active while rendering nothing — and since Manage Widgets no longer surfaces the stickies toggle, the user couldn't turn it off to recover.

Fix matches the recommendation: `DashboardWidgets` now derives `visibleWidgets` (excluding `HIDDEN_HOME_WIDGET_KEYS`) and `isAnyVisibleWidgetEnabled`, and renders over `visibleWidgets`, so hidden widgets can neither suppress the empty state nor occupy the grid.

PR #213, opened 2026-06-09
PDASHOSS01-4 add support for open claw

irichard00:
## What

Adds **OpenClaw** as a fourth runner agent backend, alongside Codex, Claude Code, and Cursor. It's selectable via `AgentKind::OpenClaw` (config `kind = "open_claw"`, CLI `--agent open-claw`) and exposes the same `AgentBridge` surface as the other three.

Resolves PDASHOSS01-4.

## How it works

OpenClaw's "app-server mode" equivalent is the **Agent Client Protocol (ACP)** — a JSON-RPC-over-stdio protocol — not a Codex-style `app-server` subcommand. The new bridge drives the headless ACP client [`acpx`](https://github.com/openclaw/acpx):

```
acpx --cwd <dir> --format json --approve-all openclaw exec -- "<prompt>"
```

`acpx` performs the ACP handshake with OpenClaw (`initialize` → `session/new {cwd}` → `session/prompt`) and streams the raw ACP JSON-RPC traffic as NDJSON on stdout. This keeps the bridge **structurally identical to the one-shot cursor-agent bridge** (prompt rides in argv, subprocess spawned lazily per turn).

Translation of ACP frames into agent-agnostic `BridgeEvent`s:
- `session/update` notifications → `Raw`, labelled by the `sessionUpdate` kind (`agent_message_chunk`, `tool_call`, `tool_call_update`, `plan`, …)
- the `session/prompt` response's `stopReason` → `Completed` (`end_turn`/`max_tokens`/…), or `Failed` for `cancelled`/`refusal`
- JSON-RPC error responses → `Failed`

## Files

- **New** `runner/src/openclaw/{schema,process,bridge,mod}.rs` — ACP NDJSON schema, one-shot subprocess wrapper, and the bridge + translation, mirroring `runner/src/cursor_agent/`.
- `AgentKind::OpenClaw` + `[openclaw]` config section (`binary = "acpx"`) in `config/schema.rs`, registered in `lib.rs`.
- Every `AgentKind` match site: `agent/mod.rs` dispatch, `daemon/supervisor.rs` failure mapping, `tui/views/config.rs` (field + section + cycle), `cli/{runner_ops,doctor}.rs`, plus the additional `RunnerConfig` initializers the compiler surfaced.
- `apps/web/.../add-runner-modal.tsx` — adds the `open-claw` option so it's selectable from the cloud add-runner UI.
- `runner/README.md` — module-tree entry.

## Design decision (A vs B)

The two prior comments on the issue asked whether to drive OpenClaw via **(A)** the `acpx` CLI or **(B)** a native ACP/JSON-RPC client against `openclaw acp`, and recommended (B). With no reply across two rounds, I proceeded with **(A)** for the MVP, because deeper research surfaced a fact that changes the tradeoff: `openclaw acp` is **not** a standalone agent — it bridges ACP-over-stdio to a running **OpenClaw Gateway over WebSocket** (requires `--url` + token), which can't be exercised in the runner env and adds real operator infra. Option (A) instead:

- maps cleanly onto the proven one-shot cursor bridge (lowest implementation risk),
- needs no live Gateway, and
- uses coarse approvals (`--approve-all`), matching the cursor MVP's already-accepted `--force` bypass posture.

**Follow-up (sequences option B, doesn't replace it):** real per-tool approval by driving `openclaw acp` natively — a persistent bidirectional ACP session answering `session/request_permission` from `runner/src/approval`. Happy to flip to (B) if preferred.

## Testing

- `cargo build` — clean
- `cargo test --lib` — **328 passed, 0 failed** (13 new openclaw schema/translation tests; no wiring regressions)
- `cargo clippy --all-targets` — clean; new `openclaw/*.rs` are rustfmt-clean

Validation is **fixture-based** because neither `acpx` nor `openclaw` is installed in the runner env (and a live OpenClaw Gateway would be needed for an end-to-end turn). This is the same constraint under which the cursor bridge shipped in #202.


irichard00:
## Automated review — OpenClaw runner backend

Reviewed the bridge/process/schema module and every `AgentKind` wiring site against the cursor-agent reference (#202). The structure is faithful to the established one-shot bridge shape, `cargo test --lib` is green (13 new fixture tests), and clippy is clean. One concrete fix applied, plus two items worth tracking.

### Fixed in this pass (pushed to the branch)
- **Missing i18n catalog entry for `OpenClaw`.** The add-runner modal calls `t("OpenClaw")`, but the key was never registered in `packages/i18n/src/locales/*/translations.ts` — unlike `Codex` / `Cursor` / `Claude Code`, which all have entries. The `t()` fallback renders the raw key so it isn't a hard failure, but it breaks the project's i18n key-sync convention (cf. `chore(i18n): sync runner detail page translation keys`) and leaves the string untranslatable in the other 18 locales. Ran `pnpm i18n:sync` to add it to the `en` source catalog plus placeholders for the rest. `@pi-dash/i18n` still type-checks.

### Worth tracking (not blocking — matches the accepted cursor MVP posture)
- **The `acpx` CLI contract is unverified.** The whole invocation — `acpx --cwd <dir> --format json --approve-all [--model <m>] openclaw exec -- "<prompt>"` emitting raw ACP JSON-RPC NDJSON on stdout — is reconstructed from docs and validated only by fixtures; neither `acpx` nor `openclaw` is installed in the runner env. This is the same fixture-based posture #202 shipped with, so it's a reasonable MVP stance, but it does need a live smoke test before it can be trusted in production. One detail I'd flag as the most likely to be wrong: `--model` is placed among **acpx's global options** (before the `openclaw exec` subcommand). If model selection is actually an `openclaw exec` option (or owned by OpenClaw config), that placement will misbehave. It's gated behind `model_default` being set (`None` by default), so default runs are unaffected.
- **Multi-turn `thread_id` instability.** Because `acpx ... exec` is stateless per turn (no `--resume`), `wait_for_session` overwrites `session_id` with a fresh ACP session id each turn, so a multi-turn chat surfaces a different `thread_id` per turn — unlike cursor, which keeps it stable via `--resume`. This is documented as a known limitation in the module header; calling it out here so reviewers can confirm the cloud's thread grouping tolerates it.

The native-ACP approach with real per-tool approvals (the documented follow-up) remains the right next step for approval fidelity.

PR #225, opened 2026-06-10
docs(design): worktree pooling — decouple runners from work dirs

irichard00:
## Summary

Design doc for worktree pooling (`.ai_design/worktree_pooling/design.md`): a daemon-managed pool of git worktrees per work dir, so N runners (e.g. a codex runner and a claude_code runner) can serve one repo from one canonical clone, with a per-work-dir local queue ("wait, don't fail") when all worktrees are leased.

Key points:

- **Pool mechanics**: lazy worktree creation up to `pool_size` (default 2), crash-safe on-disk leases, branch-lock awareness (git's one-branch-one-worktree rule), detached-HEAD parking for idle worktrees, salvage-before-clean so failed runs never lose their tree.
- **Local queue, cloud-visible**: FIFO per work dir with head-of-line bypass for branch contention only. New `WAITING_FOR_WORKTREE` run status via a new `queued` lifecycle endpoint (wire v4 HTTP); `accept` moves to lease-acquisition time; session-open redelivery rebuilds the queue after a daemon restart; cancel works while queued.
- **Matcher integration**: status joins `NON_TERMINAL_STATUSES`/`BUSY_STATUSES`; optional `free_worktrees` poll-status field as an assignment preference (hint, not gate).
- **Clean modes**: `keep-ignored` (default, warm pools) / `allowlist` / `full`, with the isolation tradeoff documented; `setup_command` failures get bounded retry then terminal `WorkspaceSetup` failure + workdir-unhealthy state.
- Two ADRs: queue lives in the daemon (with four cloud-visibility guardrails), and pool-from-day-one vs per-runner worktrees.

Companion work (separate doc, not included): agent-aware routing so issues can target a specific agent kind.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #227, opened 2026-06-10
feat: worktree pooling — decouple runners from work dirs (end-to-end)

irichard00:
Implements the full `.ai_design/worktree_pooling/` design + implementation plan end-to-end across the Rust runner, Django cloud, and web/types. Lets **N runners share one repo checkout** via a daemon-managed pool of git worktrees, with a local queue ("wait, don't fail") when all desks are busy. The motivating case — a codex runner and a claude_code runner on one repo, issues routed to a specific agent — now works.

Design + plan docs are included (first two commits). This PR supersedes the design-only PR #225.

## What's here, by phase

**Phase 1 — config (`runner/src/config/schema.rs`)**
`[[workdir]]` entity (canonical clone path, `pool_size` default 2, `clean_mode`, `keep_paths`, `setup_command`, `worktrees_dir`) and a per-runner `workdir` reference. Pooled runners are exempt from the legacy duplicate-working-dir check; work dirs get their own equal/nested path-collision checks. Hybrid with the legacy `workspace.working_dir` so existing runners are untouched.

**Phase 2 — pool library (`runner/src/workspace/{git,pool}.rs`)**
Git worktree primitives (add/remove/prune, detached-HEAD parking, mode-aware reset+clean, branch-lock listing via `worktree list --porcelain`, WIP salvage) and the `WorktreePool` **actor**: single owner task per work dir → lease/release, lazy growth to `pool_size`, FIFO queue with **branch-lock head-of-line bypass**, salvage-before-clean on abnormal exit, crash-safe on-disk lock reaping at startup, and workdir-unhealthy on setup failure. 13 unit tests against real git repos.

**Phase 3 — supervisor wiring (`runner/src/daemon/supervisor.rs`)**
One pool per work dir at startup; handle threaded through `RunnerLoop` → `AssignWorker`. `handle_assign` leases a worktree (racing the cancel signal so a cancel-while-queued dequeues), uses it as the agent cwd, releases on every exit path, marks success only on clean completion. Legacy runners unchanged.

**Phases 4–5 — cloud status + redelivery (`apps/api/pi_dash/runner/`)**
`WAITING_FOR_WORKTREE` status + `RunQueuedEndpoint` (`runs/<id>/queued/`, `ASSIGNED→WAITING`, no regression from RUNNING/terminal); status joins `NON_TERMINAL_STATUSES`/`BUSY_STATUSES`; cancel works from waiting; `queue_position` on `AgentRun`; session-open `redeliver` payload rebuilds a restarted daemon's queue. Migration `0017_worktree_pooling`.

**Phase 6 — protocol flip (`runner/src/cloud/{protocol,http}.rs`)**
`ClientMsg::RunQueued` dispatched to the `queued` verb with **404 feature-detection** — an old cloud latches a per-session flag and the runner silently falls back to accept-on-enqueue. `handle_assign` emits `RunQueued` when a pooled run parks.

**Phase 7 — capacity hint**
`free_worktrees` reported in the long-poll status body (from the pool snapshot); `select_runner_in_pod` prefers a runner with a free desk (preference, never a gate).

**Phase 8 — surface**
`pidash workdir add|list|remove`; `pidash runner add --workdir`; `StatusSnapshot` carries per-workdir pool occupancy/queue/health, rendered by `pidash status`. Web: `waiting_for_worktree` threaded through `TAgentRunStatus`, exhaustive status maps, the issue agent-status panel, and the runs page (+ i18n).

## Verification
- **Rust:** 355 lib tests + integration tests pass; `cargo clippy --all-targets -D warnings` clean.
- **Django:** 38 new pooling/matcher tests pass against real Postgres; 92 existing runner tests pass (no regressions).
- **Web:** `@pi-dash/types` + changed web files type-check clean.


irichard00:
## Adversarial review pass (4 independent reviewers + cross-stack contract check) — all findings fixed in b4c57275

**Critical / High (would have broken the feature in production):**
1. **Daemon restart bricked the pool** — leftover \`wt-N\` dirs from clean shutdowns collided with \`git worktree add\`; init now adopts every leftover desk (salvage → clean → park → register), removing corrupt ones.
2. **Redelivery was dead end-to-end** — the cloud built the session-open \`redeliver\` payload but the runner never parsed it, *and* the session-open reaper failed any ASSIGNED/WAITING run older than 60s before redelivery could fire. Both sides fixed; the Django regression test now uses a 10-minute-old run.
3. **Branch double-grant** — \`git checkout -B\` does not enforce one-branch-one-checkout, and the actor's pre-check treated git errors as "unlocked". Now fail-closed at both layers.
4. **Eternal queue on a broken branch** — checkout failures now count against a bounded retry (with a scheduled re-drain), instead of parking forever on an idle pool.
5. **Duplicate runs while queued** — \`WAITING_FOR_WORKTREE\` was missing from \`_active_run_for\`, the bgtasks scheduler's non-terminal set, and the issue serializer's \`active_run\` filter; ticks could dispatch a second concurrent run and the issue panel never saw the queued state.
6. **Chat sessions bypassed the pool** — they ran in the vestigial \`working_dir\` (or the canonical clone); they now hold a \`LeaseKind::Session\` desk for the session lifetime.
7. **Lost cancels** — the edge-triggered \`Notify\` dropped a Cancel arriving before the worker registered; replaced with a level-triggered \`CancellationToken\`.

**Medium:** salvage now covers branch-less aborts and survives the next lease of the branch (no more \`-B\` force-reset over unpushed salvage commits); queue positions are reported authoritatively by the pool owner (on park + every change) instead of a racy pre-acquire snapshot; the \`queued\` 404 latch only trips on route-shape 404s, resets per session, and latched runners fall back to accept-on-enqueue per design §15; \`queue_position\` is sent to the issue UI and cleared on accept/requeue/terminal; setup_command is time-bounded; unhealthy pools advertise 0 free desks; validation now rejects legacy-runner/work-dir overlaps and worktrees_dir nesting.

**Verification:** 367 Rust lib tests + all integration suites, \`clippy --all-targets -D warnings\` clean; 326 Django runner tests + 70 pooling/drain/orchestration tests green in the live Postgres container (\`--create-db\`); web types clean. New regression tests cover every Critical/High finding.

PR #231, opened 2026-06-11
fix(prompting): route PR-producing runs to In Review, not Done

irichard00:
## Problem

When an agent finishes a coding task and opens a PR, the run was ending by moving the issue straight to a **`completed`-group state ("Done")**, skipping **"In Review"**. At that point the PR is awaiting human review and merge — marking the issue Done falsely claims the change already landed.

Observed on a local Codex run (`e964bb28`, issue `HOMEP001-2`): the agent correctly opened the PR, then ran `pidash issue patch HOMEP001-2 --state "Done"`.

## Root cause: prompt design, not agent disobedience

The orchestration prompt steered every successful run to the `completed` group and gave `--state "Done"` as the **only** canonical success example, across four fragments:

- `03_pidash_cli.md` "Typical recipes" — the worked example was `--state "Done"`
- `09_implementation.md` Step 2.8 — "move the issue to a state in the `completed` group"
- `13_ending_run.md` §1 — "a state in the `completed` group (usually Done)"
- `04_default_posture.md` — same framing

Comparing two agents on the same (agent-agnostic, server-generated) prompt made the gap obvious:

- **Codex** took the worked example literally and ran `--state "Done"`.
- **Claude Code** overrode it to **In Review** on its own judgement — its run log even says Done "would falsely imply [completion]" for an open PR.

So the prompt was coasting on the agent being willing to second-guess it. This removes the ambiguity instead.

## Fix

Route the final state by task type, in all four fragments:

- `code_change` that opened a PR → **`review` group ("In Review")**
- finished `noncode` task with no PR → `completed` group ("Done")
- fall back to `completed` only if the project exposes no `review`-group state

## Test

Adds `test_pr_work_routes_to_review_group_not_done` (in `tests/unit/prompting/test_fragments.py`): asserts the assembled prompt offers the In Review move and presents it ahead of the Done example. Existing fragment-assembly tests (section ordering, CLI subsections, Jinja block-tag column rule) are unaffected — verified by simulating `assemble()` locally.


irichard00:
Superseded by #250. This PR targeted the old `prompting/fragments/*` system, which #235 replaced with section-based prompt management (`prompting/sections/*` + `recipes.py`) — the fragments are now dead code, so this branch conflicts and can't be reapplied in place. #250 reapplies the identical fix (route PR-producing runs to In Review, not Done) to the live section files, with the regression test ported into `test_composer.py`.

PR #232, opened 2026-06-11
feat(web): independent scroll + async detail loading on runners/runs

irichard00:
## What

Improves the UX of `/<workspace>/runners/runs`, resolving **PDASHOSS01-24**:

1. **Async detail loading** — the right-hand run-detail pane no longer makes the page wait on a fully-loaded run.
2. **Independent scrolling** — the run list (left) and run detail (right) now scroll independently instead of as one page-level container.

## How it works

- **Independent scroll.** The page is constrained to the viewport (`h-full min-h-0` flex column) and the two-pane grid gets `flex-1 min-h-0 overflow-hidden`. Each pane becomes its own `overflow-auto` scroll region. The list table header is `sticky top-0` so column labels stay visible while the list scrolls.
- **Async / non-blocking detail.** Detail already fetched async via SWR, but while a run was selected and still loading, the pane showed the misleading **"Select a run on the left."** empty state. That branch is now three-way: no selection → prompt; selected but loading → `Spinner` + **"Loading run details…"**; loaded → detail. The page never blocks on detail load.

## i18n

Registered the new `"Loading run details…"` key via `pnpm i18n:sync` — English source value plus empty placeholders for the other 18 locales (matching the project's established key-sync process).

## Files

- `apps/web/app/(all)/[workspaceSlug]/runners/runs/page.tsx` — scroll containers + three-state detail pane.
- `packages/i18n/src/locales/*/translations.ts` — new `"Loading run details…"` key across all locales.

## Testing

- `oxlint --deny-warnings` + `oxfmt --check` on the changed page and en catalog → clean.

### DIFF
--- a/apps/web/app/(all)/[workspaceSlug]/runners/runs/page.tsx
+++ b/apps/web/app/(all)/[workspaceSlug]/runners/runs/page.tsx
@@ -14,7 +14,7 @@ import { RunnerService } from "@pi-dash/services";
 import type { IAgentRun, TAgentRunStatus } from "@pi-dash/types";
 import { AGENT_RUN_TERMINAL_STATUSES } from "@pi-dash/types";
 import type { TBadgeVariant } from "@pi-dash/ui";
-import { AlertModalCore, Badge, Button } from "@pi-dash/ui";
+import { AlertModalCore, Badge, Button, Spinner } from "@pi-dash/ui";
 import { PageHead } from "@/components/core/page-title";
 import { RunnersTabs } from "@/components/runners/runners-tabs";
 import { useWorkspace } from "@/hooks/store/use-workspace";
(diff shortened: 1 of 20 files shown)


irichard00:
## Review — PDASHOSS01-24 (automated review pass)

Verified both acceptance criteria are met and the change is correct:

- **Independent scroll** ✓ — page root `h-full min-h-0` flex column, grid `flex-1 min-h-0 overflow-hidden`, each pane its own `overflow-auto`. The height chain holds: `runners/layout.tsx` renders `<main className="flex-1 …">` inside a `flex h-full` row, so `main` has a definite height and `h-full` on the page resolves correctly.
- **Async / non-blocking detail** ✓ — the three-way branch (no selection → prompt; selected & `!detail` → Spinner; loaded → detail) is correct, and SWR keeps `data` during background revalidation so the spinner only shows on first load, not on every 3s refresh.

No correctness issues found.

### Re: the prior automated suggestion to guard `detail.id !== selected`

This isn't needed and would be dead defensive code. The detail hook is keyed by the selected run — `useSWR(["runner-run-detail", selected], …)` — so `detail` is read from the cache slot for the *current* `selected`. When selection changes, SWR returns either the new key's cached value (which is that run's data) or `undefined` (→ spinner). `keepPreviousData` is not enabled locally or in the global `WEB_SWR_CONFIG`, so `detail` can never hold a previously-selected run's data while a different run is selected. The "stale data flash" the guard targets cannot occur here.

🤖 Automated review (Claude)

PR #245, opened 2026-06-12
feat(prompting): tell the agent what triggered its run and the ticking schedule

irichard00:
## Problem

Every continuation run starts a fresh agent session (by design — no session resume). The agent reconstructs *what to do* from the issue description, workpad, and comments, but it has no way to know **why this particular run fired**: a scheduled ticker re-invocation, a new human comment, a manual "Run AI" click, and a state transition all produce an identical prompt. The `triggered_by` value existed internally (creator resolution + logs) but never reached the prompt.

That blind spot means a 3-hourly idle tick gets the same treatment as an explicit human nudge — the agent can't choose to exit cheaply when nothing has changed, or prioritize the new comment when one arrived.

## Change

**Backend wiring**
- `_create_continuation_run` / `_create_and_dispatch_run` now take `triggered_by` and persist it in `AgentRun.run_config["triggered_by"]` *before* the prompt renders. All four dispatch paths pass it: tick (`tick`), comment continuation (`comment` / `comment_and_run`), Run AI (`run_ai`), and state transitions (`state_transition`, the default).

**Prompt context** (`build_context`)
- `run.trigger` — the persisted trigger, `None` for pre-existing runs.
- `tick` — the issue's ticking schedule from `IssueAgentTicker`: `count`, `cap`, `remaining`, `interval_seconds`, `interval_human`. `cap`/`remaining` surface as `None` for the infinite (`-1`) cap; `tick` is `None` when no ticker row exists.

**Templates**
- New **"Why this run started"** section in `02_session_framing.md` with per-trigger guidance. Tick runs: diff the workpad/comments against reality, don't redo recorded work, exit promptly if nothing is actionable. Comment runs: address the new comment first. Plus a ticking-schedule line ("re-invoked about every 3 hours, used 5 of 24 ticks, 19 remaining before auto-pause").
- Same block in the review template; an unchanged work product on a tick should emit `noop` instead of re-reviewing.
- Migration `prompting/0004` force-reseeds both template bodies (the DB row is the runtime source of truth; follows the `0003` pattern).

## Tests

- `build_context`: trigger surfaced / `None` fallback, tick block populated, infinite-cap → `None`, no-ticker → `None`.
- Fragment renders for tick / comment / unrecorded triggers.
- End-to-end: `fire_tick` → run has `triggered_by == "tick"` and the rendered prompt contains the ticker guidance.
- `pi_dash/tests/unit/{prompting,orchestration,bg_tasks/test_agent_ticker.py}`: **141 passed**. Ruff clean. (Remaining unit-suite failures are pre-existing on main — verified by stashing this change.)


irichard00:
## Code review follow-up

A multi-angle review of this PR surfaced 4 confirmed correctness issues, all fixed in cd506fa2:

| Finding | Fix |
|---|---|
| Template-preview endpoint 500'd (`_FakeRun` has no `run_config`; 2 preview contract tests were failing) | `build_context` reads `run_config` via `getattr`; preview tests green again |
| Tick-schedule prose rendered for **disarmed** tickers — false "automatically re-invokes" promise that invites the agent to defer work to a tick that never fires | `_tick_context` returns `None` unless `ticker.enabled`; also omits the block for nonsense config (interval ≤ 0, negative non-infinite cap — reachable via the unvalidated project-default fields) |
| Comment & Run / Run AI baked the **stale pre-reset budget** into the prompt ("used 23 of 24") then refunded it in the same request | Reset moved before dispatch (matching `handle_issue_comment`'s order), with transaction rollback on dispatch failure so a 409 still doesn't refund budget — both paths covered by new regression tests |
| A human comment **coalesced** into a QUEUED tick run executed against "not a human action … exit promptly" prose | Both templates now flag the inlined thread as a creation-time snapshot and direct the agent to `pidash comment list` before concluding nothing changed |

**Surfaced but deliberately deferred** (follow-up sized, not blockers):
- `triggered_by` lives in `run_config` (JSON grab-bag, client-writable via POST /runs) rather than a first-class `AgentRun` CharField with choices — the design doc (§7.4 issue_ticking_system) deferred that field "until a use case appears"; this PR arguably is the use case.
- Trigger labels split across two modules + hardcoded in two templates; a `TextChoices` would be the single source of truth.
- The "Why this run started" block is hand-duplicated between the fragment and the review template (two Jinja styles); seed-time composition could share one copy.
- Migration 0004 follows the 0003 precedent (live-model imports, force-reseed of global rows) — convention, but worth revisiting if global templates ever become operator-owned.
- Workspaces with pre-existing template overrides won't get the new section until they update their override (by design of the override mechanism).

irichard00:
## Merged latest main + reconciled with two overlapping changes

`main` moved 11 commits since this branch was cut, including two that directly overlap this PR:

1. **#235 — section-based prompt management.** The live prompt is now composed from `prompting/sections/*.md` via `recipes`/`composer`, **not** the old `fragments/ → PromptTemplate` path. My edits to `fragments/02_session_framing.md` and `REVIEW_TEMPLATE_BODY` were therefore dead on main.
2. **`AgentRun.trigger` field** (`AgentRunTrigger` choices, `db_index=True`, `HUMAN_TRIGGERS`/`run_is_human_triggered`) — a first-class trigger field, exactly the altitude fix the earlier review of this PR recommended. `orchestration/{service,scheduling}.py` already thread it through every dispatch path.

### What changed in the reconciliation
- **Dropped** my entire `run_config["triggered_by"]` plumbing — `context.py` now reads main's `run.trigger` field (`getattr` for preview-stub safety). Took main's `service.py`/`scheduling.py` wholesale.
- **Relocated** the "Why this run started" + ticking-schedule guidance into the shared `sections/session-framing.md`, gated `{% if run.kind != "scheduler" %}` (the established pattern; also keeps scheduler runs `StrictUndefined`-safe, since their context has no `trigger`/`tick`). One copy now serves coding **and** review (with a review-specific `noop` hint) — this also resolves the earlier duplication finding.
- **Added** `trigger`/`tick` to `validation.sample_contexts` so the full-prompt validation contract (#235) covers the new section references.
- **Dropped** my reseed migration — collided with main's `0004_prompt_section_override`, and the section system reads defaults from code (no seed step). Migration graph now has a single leaf per app.
- **Kept** all three bug fixes from the review: `_tick_context` gates on `enabled` + sane config; reset-before-dispatch in the two button endpoints (survived the merge); snapshot-aware tick prose.
- Tests ported from the deleted `test_fragments.py` into `test_composer.py` (section render: coding tick / review noop / comment / scheduler-omits-block) + `test_context.py`.

### Verification
`mergeable: MERGEABLE` (conflicts resolved). Affected suites green against merged main: **212 passed** (prompting unit+contract, agent_ticker, orchestration, runner trigger) + **27 passed** (runs-views). Two `test_runs_views` failures (`*_skip_immediate_dispatch_*`) are **pre-existing on main** — confirmed by reverting my source files to main and re-running; they fail identically with `[Errno 111] Connection refused` in the immediate-dispatch path (environmental, not prompt-related). Ruff clean.

Net diff vs `main` is 8 files: `context.py`, `sections/session-framing.md`, `validation.py`, `runner/views/runs.py`, and four test files.

PR #271, opened 2026-06-22
[codex] Add Theme F platform federation

irichard00:
## What changed
- Added platform identity fields for users, workspaces, and workspace members plus webhook delivery/federation state models and migration.
- Added signed IOS webhook ingestion with replay protection, event version ordering, owner bootstrap, member add/role/revoke handling, and runtime access revocation.
- Added platform launch-token consumption at `/auth/platform/session/` with local session creation.
- Added reconciliation task support for rebuilding linked workspace/member state from Home Page's internal IOS API.
- Guarded linked workspaces from local invite membership changes and added disabled-workspace checks to shared permission paths.
- Registered Theme F platform configuration keys.
- Added focused contract tests for signing, idempotency, stale ordering, revocation cleanup, and session consumption.

## Why
Pi Dash has its own database, so it needs a local projection of central enterprise identity plus signed event delivery and reconciliation instead of direct data_store access.

## Validation
- `DATABASE_URL=postgresql://... ./.venv/bin/python -m pytest pi_dash/tests/contract/app/test_platform_federation.py -q` passed against local Docker Postgres.
- `git diff --check` passed.

Note: `makemigrations --check --dry-run --skip-checks` still reports unrelated existing model/migration drift in `db` and `runner`; Theme F fields are covered by `0152_platform_federation.py`.


irichard00:
Superseded by the private-pi-dash implementation in The-AI-Republic/private-pi-dash#77. Theme F should live in the private overlay, not the OSS repo.

PR #279, opened 2026-07-02
PDASHOSS01-41 new user has to create workspace?

irichard00:
## What & why

Issue: **PDASHOSS01-41 — new user has to create workspace?**

Onboarding currently forces every uninvited new user into "Create your workspace" with no way to join an existing one. Per the discussion on the issue, the chosen flow is **request-to-join by admin email**: a signed-in but workspace-less user types a workspace *admin's email* to ask to join, and a workspace admin approves or denies. While waiting, the user intentionally stays on the onboarding page (a "pending approval" step — frontend, next PR).

This is the **first of three PRs** — this one is the backend only:

1. **Backend** (this PR) — join-request model + endpoints.
2. Requester onboarding UX — the "join by admin email" entry in the create step + the new "pending approval" step.
3. Admin approval UI (Members → Requests) + notification.

## Changes

- **`WorkspaceJoinRequest` model** (migration `0153`) — the inverse of `WorkspaceMemberInvite`. `workspace` is **nullable on purpose**: when the typed email is nobody's admin, an unresolved request is still recorded so the requester's onboarding "pending" state is identical whether or not the email was real — this avoids leaking which emails are workspace admins (enumeration). A partial unique constraint prevents duplicate *pending* requests to the same workspace.
- **Requester endpoints** under `/api/users/me/workspaces/join-requests/` (reachable without a workspace slug):
  - `POST` — create a request by `admin_email`. Resolves the email to the workspace(s) it administers (active Admin member or owner), skips workspaces the requester is already in, and is idempotent. **Always** returns a neutral `{"message": "Request sent"}`.
  - `GET` — list the current user's own requests (drives the pending step in PR 2).
- **Admin endpoints** under `/api/workspaces/<slug>/join-requests/` (`WorkSpaceAdminPermission`):
  - `GET` — list pending requests for the workspace.
  - `POST .../<pk>/approve/` — create/reactivate the `WorkspaceMember`, point the requester's `Profile.last_workspace_id` at the workspace, mark `APPROVED`.
  - `POST .../<pk>/deny/` — mark `DENIED`.

## Testing

- 13 new contract tests in `test_workspace_join_request_app.py` (resolution, enumeration-neutrality, invalid/own email, idempotency, requester list scoping, admin list, approve creates membership, double-approve rejection, deny, non-admin denied). All pass against Postgres + valkey.
- `manage.py check` clean; migration `0153` is in sync with the model and applies cleanly on a fresh DB; `ruff check` / `ruff format` clean on the new files.

## Notes for the reviewer

- **Role granted on approval** defaults to Member (15); the admin approval gate is the real control. Easy to change.
- Discovered (not fixed here — flagging for a separate cleanup): `last_workspace_id` lives on **`Profile`**, but the existing invite endpoint (`app/views/workspace/invite.py`) sets it on `User`, which is a no-op. This PR sets it correctly on `Profile`.
- Admin notification (email/in-app) is intentionally deferred to PR 3, alongside the admin UI where it belongs.


irichard00 (commented):
## Automated review — backend join-request (PR 1 of 3)

Reviewed the model, views, serializer, URLs, migration wiring, and the 13 contract tests. The shape is sound: the enumeration-neutral design (nullable `workspace`, always-neutral "Request sent") is implemented consistently, requester vs. admin surfaces are cleanly split, and the 404/IntegrityError handling matches the existing invite viewset's patterns. Test coverage is real and matches the claims. Nothing here blocks merge — below are notes for you to weigh, several of which touch PR 2's behavior.

### 1. Members — not just admins — can approve/deny (medium, confirm intent)
`WorkspaceJoinRequestViewSet` uses `WorkSpaceAdminPermission`, which admits `role ∈ [Admin(20), Member(15)]`. So any active *member* can approve or deny join requests. This is consistent with the existing `WorkspaceInvitationsViewset` (same permission), so it may be deliberate — but the issue thread and the model docstring both frame this as "a workspace **admin** approves." If admin-only was intended, `WorkspaceOwnerPermission` (role == Admin) is the right gate. Worth an explicit decision since it's a member-management privilege.

### 2. Re-approving a deactivated member can silently demote them (low)
`join_request.py` approve path:
```python
workspace_member = WorkspaceMember.objects.filter(workspace=workspace, member=requester).first()
if workspace_member is not None:
    workspace_member.is_active = True
    workspace_member.role = join_request.role   # <- default Member(15)
```
If the requester was previously an Admin who got deactivated, reactivating via a join request overwrites their role down to Member. Consider preserving the existing role on reactivation (only set role when creating a fresh membership), or at least not downgrading.

### 3. Already-a-member of the resolved workspace → spurious null-workspace request (low/medium — affects PR 2)
In `create`, `already_member_ids` is subtracted from `target_workspace_ids`. If the typed email resolves *only* to workspace(s) the requester already belongs to, the set becomes empty and control falls into the `else` branch, which records an **unresolved (`workspace=None`) pending request** — indistinguishable from "email matched no admin." That request can never be approved, so in PR 2 it risks stranding the user in the pending onboarding step forever. Recommend distinguishing "resolved, but requester is already a member" (return neutral success, create nothing) from "did not resolve at all."

### 4. Null-workspace uniqueness is app-level only / TOCTOU (low)
The partial unique constraint is on `(requester, workspace)` with `status=PENDING`. In Postgres, `NULL` workspaces are distinct, so the constraint does **not** prevent duplicate unresolved requests — that case relies solely on the `.exists()` check, which is racy under concurrent submits. Separately, a genuine race on the resolved path would raise `IntegrityError` → the base handler returns 400 ("payload is not valid"), breaking the neutral "Request sent" contract for that caller. Both are low-probability; a `get_or_create` (or catching IntegrityError and returning the neutral response) would tighten it.

### Nits
- `approve` calls `str(timezone.now())` twice (once for `responded_at`, once in the event payload) — harmless, minor.
- `WorkspaceJoinRequestSerializer` uses `fields = "__all__"`; safe here because create bypasses the serializer and only `list` is routed, but a future PATCH/PUT route would expose the writable fields — keep in mind for PR 3.

Good to proceed to PR 2 on this base once #1 and #3 are decided.

PR #284, opened 2026-07-02
PDASHOSS01-42 Is AgentRun card reflecting the manual trigger run and comment and run?

irichard00:
## What & why

Issue PDASHOSS01-42: manually triggering a run via **Run AI** or **Comment & Run** did not update the AgentRun status card on the issue detail page — the run only showed up after a manual reload.

### Root cause

The backend was never the problem. Both buttons `POST /api/runners/runs/` → `scheduling.dispatch_run_ai_run` / `dispatch_continuation_run`, which create a full `AgentRun` row, and the card's serializer (`IssueDetailSerializer.get_agent_status`) queries **all** runs on the issue with no trigger filter. The data is there.

The gap is purely frontend:

- After `triggerRun` succeeds, nothing re-fetched the issue, so the store's `issue.agent_status` stayed stale.
- `IssueAgentStatusPanel`'s 15s live poller only starts once `agent_status.active_run` is already an active status in local state — which never became true without a refresh.

So the newly-created run stayed invisible until the user reloaded the page.

### Fix

Re-fetch the issue detail after a successful trigger. The new run lands as `queued` (an `ACTIVE_RUN_STATUS`), so the card renders it immediately and the poller starts tracking it to completion. `projectId` is threaded through the two trigger buttons so the shared `useCreateAgentRun` hook can call `fetchIssue`. The re-fetch is best-effort — a refresh failure won't turn a successful trigger into an error toast; the card catches up on its next scheduled refresh.

### Files

- `use-create-agent-run.ts` — accept `projectId`, `fetchIssue` after success
- `run-ai-button.tsx`, `action-buttons.tsx` — thread `projectId`
- `comment-and-run-button.tsx`, `issue-activity/root.tsx` — thread `projectId`

## Validation

- `pnpm turbo run check:types --filter=web` — passes
- `pnpm turbo run check:lint --filter=web` — 0 errors
- Reasoning trace of the refresh path (no runner available in-session for a live UI test). Suggested manual check: trigger a run on an idle issue and confirm the card flips to "Working"/"Queued" without reloading.

### DIFF
--- a/apps/web/core/components/issues/issue-detail-widgets/action-buttons.tsx
+++ b/apps/web/core/components/issues/issue-detail-widgets/action-buttons.tsx
@@ -93,7 +93,7 @@ export function IssueDetailWidgetActionButtons(props: Props) {
           issueServiceType={issueServiceType}
         />
       )}
-      <RunAIActionButton workspaceSlug={workspaceSlug} issueId={issueId} disabled={disabled} />
+      <RunAIActionButton workspaceSlug={workspaceSlug} projectId={projectId} issueId={issueId} disabled={disabled} />
       <WorkItemAdditionalWidgetActionButtons
         disabled={disabled}
         hideWidgets={hideWidgets ?? []}
--- a/apps/web/core/components/issues/issue-detail-widgets/run-ai/comment-and-run-button.tsx
+++ b/apps/web/core/components/issues/issue-detail-widgets/run-ai/comment-and-run-button.tsx
@@ -16,6 +16,7 @@ import { useCreateAgentRun } from "./use-create-agent-run";
 
 type Props = {
   workspaceSlug: string;
+  projectId: string;
   issueId: string;
   /** Posts the comment currently typed in the composer and returns it (or
    * ``undefined`` when empty / on failure). Supplied by ``CommentCreate`` via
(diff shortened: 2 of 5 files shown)


irichard00 (commented):
**Automated review — approved.**

Verified the fix end-to-end against the code:

- **Root cause is real and correctly diagnosed.** `agent-status.tsx` starts its 15s poller only when `hasActiveAgentRun` is true, and that derives from `issue.agent_status.active_run` in local state. With no re-fetch after a manual trigger, a freshly-`queued` run stays invisible until reload.
- **The data was always available.** `IssueDetailSerializer.get_agent_status` filters `AgentRun` by `work_item` only (no trigger filter) and includes `QUEUED` in `active_run`, so the re-fetch surfaces the new run immediately and the poller takes over.
- **Signatures/wiring check out.** `fetchIssue(workspaceSlug, projectId, issueId)` matches `issue.store.ts`; `projectId` is in scope at both call sites; both `RunAIActionButton` and `CommentAndRunActionButton` are the only callers and both are threaded correctly.
- **Failure handling is sound.** The re-fetch is best-effort (try/catch), so a refresh error can't turn a successful trigger into an error toast — the card catches up on the next scheduled poll.

CI green (test + plan). No issues found; no changes requested.

PR #287, opened 2026-07-02
PDASHOSS01-37 Pidash TUI improvement: somehow runner now shows every agent even it is not relevant

irichard00:
## Summary

Fixes the runner TUI Settings panel showing **every** agent's configuration (Codex, Claude Code, Cursor Agent, OpenClaw) regardless of which agent a runner actually drives. Now only the section matching the runner's `agent.kind` is shown — and only its fields are keyboard-navigable.

Resolves **PDASHOSS01-37**.

## The bug

In the Runners tab → Settings panel, all four per-agent sections were rendered unconditionally, and all their fields were registered in the focus tree, so `↑/↓` also stepped through the irrelevant agents' `binary` / `model_default` fields.

## The fix

Filter both the render layer and the focus tree by the picked runner's selected agent kind, using one shared predicate.

- **`runner/src/tui/views/config.rs`** — new `field_agent_kind()` (maps an agent-specific `FieldId` to its `AgentKind`) and public `field_visible()`. `editable_lines()` now draws only the section whose kind equals `runner.agent.kind`.
- **`runner/src/tui/views/runner_status.rs`** — `focus_tree()` excludes hidden agent fields so navigation skips them. Rows keep their original `FIELDS` index; `next_sibling` sorts+dedups rows, so the gaps navigate correctly. Highlight alignment is preserved because `focused_field_idx` resolves the leaf by `id_str` position, independent of tree rows.
- **`runner/src/tui/app.rs`** — `dispatch_focus_key()` gains a stale-leaf guard that pops focus segments no longer present in the current tree. This handles the edge case of switching runners across agent kinds while dived into an agent-specific field (the field the focus pointed at can disappear).

`differs()` is intentionally left untouched — it compares all `FIELDS` for unsaved-change detection, which is correct regardless of visibility.

## Behavior

- Cycling **Agent → kind** (Enter on the always-visible `kind` field) immediately swaps which agent section is shown.
- Editing the visible agent's `binary` / `model_default` still works and saves as before.

## Tests

New unit tests:
- `focus_tree_hides_non_selected_agent_fields`
- `focus_tree_agent_fields_follow_selected_kind`
- `editable_lines_shows_only_selected_agent_section`
- `field_visible_matches_selected_kind`

`cargo build -p pidash` is clean and `cargo test -p pidash --lib tui::` passes. (One pre-existing, environment-specific failure in `workspace::git` — a macOS `/private/var` vs `/var` tmpdir symlink mismatch — also fails on `main` and is unrelated to this change.)

### DIFF
--- a/runner/src/tui/app.rs
+++ b/runner/src/tui/app.rs
@@ -805,6 +805,19 @@ impl App {
             return KeyHandled::NotConsumed;
         }
 
+        // Stale-leaf guard: a tab's focus tree can shrink between frames — e.g.
+        // switching to a runner whose `agent.kind` hides the agent field the
+        // user had dived into. Pop any trailing segments that no longer resolve
+        // in the current tree so navigation never strands on a leaf that isn't
+        // there; the focus settles on the nearest still-present ancestor.
+        loop {
+            let segs = self.focus_for(self.tab).segments().to_vec();
+            if segs.is_empty() || locate(&tree, &segs).is_some() {
+                break;
+            }
+            self.focus_for_mut(self.tab).pop();
+        }
+
         let focus = self.focus_for(self.tab).clone();
         let active_ctxs = self.active_tab().active_contexts(&focus);
         let text_input_only = active_ctxs == [Context::TextInput];
--- a/runner/src/tui/views/config.rs
+++ b/runner/src/tui/views/config.rs
@@ -200,6 +200,41 @@ pub fn field_at(idx: usize) -> &'static FieldSpec {
     &FIELDS[idx.min(FIELDS.len() - 1)]
 }
 
+/// The `AgentKind` an agent-specific config field belongs to, or `None`
+/// for fields that aren't tied to a particular agent (those are always
+/// shown regardless of the runner's selected kind).
+pub fn field_agent_kind(id: FieldId) -> Option<AgentKind> {
+    match id {
+        FieldId::CodexBinary | FieldId::CodexModelDefault => Some(AgentKind::Codex),
+        FieldId::ClaudeBinary | FieldId::ClaudeModelDefault => Some(AgentKind::ClaudeCode),
+        FieldId::CursorBinary | FieldId::CursorModelDefault => Some(AgentKind::CursorAgent),
+        FieldId::OpenClawBinary | FieldId::OpenClawModelDefault => Some(AgentKind::OpenClaw),
+        _ => None,
+    }
+}
+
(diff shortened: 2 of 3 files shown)


irichard00 (commented):
## Automated review — PDASHOSS01-37 ✅

Reviewed the diff (3 files, +253/−5) and validated it against the acceptance criteria. **No blocking issues found; the fix is correct and well-tested.**

### What I verified

- **Single shared predicate.** `field_visible()` in `config.rs` is the one source of truth for "is this field shown for this runner", keyed off `runner.agent.kind`. Both the render layer (`editable_lines`) and the focus tree (`runner_status.rs::focus_tree`) consume it, so what's drawn and what's navigable can't drift apart.
- **Consistent runner index.** Render (`runner_status.rs:486`) and focus tree (`runner_status.rs:119`) both resolve the index via `data.picker_runner_index().unwrap_or(0)` against `config_working`, and `field_visible`/`editable_lines` both index `cfg.runners` through `runner_at`. Same runner, same config, same result.
- **Navigation across the index gaps.** Hidden fields are dropped from the focus tree but visible rows keep their original `FIELDS` index. `next_sibling` (Up/Down) collects rows, `sort_unstable` + `dedup`, then steps to the adjacent present row — so the non-contiguous indices navigate cleanly. Confirmed in `view/focus.rs:255-275`.
- **Highlight stays aligned.** `focused_field_idx` resolves the selection by `id_str` position in the full `FIELDS` table, independent of the filtered tree, so the render highlight lines up with the focused visible row.
- **Stale-leaf guard.** `app.rs::dispatch_focus_key` pops trailing focus segments that no longer `locate()` in the current tree — covers switching runners across agent kinds while dived into an agent field. The loop is bounded (each iteration `pop()`s; terminates on empty or resolved), no infinite-loop risk.
- **`differs()` correctly left untouched** — it compares all `FIELDS` for the unsaved-change marker, which must stay independent of visibility.

### Validation

Ran `cargo test -p pidash --lib tui::` on the PR branch in a clean worktree: **43 passed, 0 failed**, including the 4 new tests (`focus_tree_hides_non_selected_agent_fields`, `focus_tree_agent_fields_follow_selected_kind`, `editable_lines_shows_only_selected_agent_section`, `field_visible_matches_selected_kind`). Build is clean.

Only render surface for the agent sections is this Runners-tab Settings panel (`editable_lines` has a single caller), so the fix fully covers the reported bug. Ready to merge.

PR #293, opened 2026-07-02
PDASHOSS01-34 Make Pi Dash Cloud AgentRun have script like user  use the agent cli

irichard00:
## Problem

Pi Dash AgentRun records only captured lifecycle and error events (e.g. "Run failed. 401 Invalid authentication credentials"). The agent's human-readable narrative — the *"let me first explore the original color of the home page button…"* prose a user sees in their terminal when running the agent CLI directly — never made it into the records.

The runner folded those `assistant/message` frames into a **count-only** compaction summary (`{"summary": "N low-signal agent events", ...}`) and dropped the actual text.

Closes PDASHOSS01-34.

## Change (runner-side only)

Forward the agent's narrative to the cloud **verbatim** as its own `agent/message` run event, so an AgentRun reads like a live CLI session:

- **`observability.rs`** — `extract_agent_message_text(method, params)` reads the visible narrative from:
  - Claude `assistant/message` → concatenated `text` content blocks (skips `thinking`/`tool_use`).
  - Codex `item/completed` → `agentMessage` item `text`.

  Trimmed and length-capped (`AGENT_MESSAGE_TEXT_MAX = 4000`).
- **`run_event_mirror.rs`** — `push_content(kind, payload)` records a full-content event verbatim, flushing the compact buffer first so ordering relative to surrounding low-signal events is preserved.
- **`supervisor.rs`** (`handle_bridge_event`) — routes narrative Raw frames to `push_content("agent/message", {schema: "runner_agent_message_v1", text})`. Lifecycle milestones and high-frequency token deltas (`stream_event/*`, `*/delta`) keep the sanitised-summary / compaction path, so mirrored event volume stays bounded.

## No API/frontend change needed

The cloud `RunEventEndpoint` already persists arbitrary (byte-capped) payloads into `AgentRunEvent`, and `AgentRunEventSerializer` exposes `payload` — so the narrative is available via the run-detail endpoint (`?include_events=1`) with no backend change.

**Follow-up (out of scope):** no web UI currently renders `AgentRunEvent` narrative — a run-transcript view would surface this to users. The issue only required the *records* to contain the data, which this delivers.

## Validation

- `cargo test --lib observability::tests::extract_agent_message` → 7 passed
- `cargo test --lib run_event_mirror` → 9 passed (incl. new content-forwarding test asserting prior deltas compact, then the message forwards verbatim)
- `cargo test --lib daemon::` → 77 passed (no regressions)
- `cargo clippy --lib` → clean


irichard00 (commented):
## Automated review — approved ✅

Reviewed the diff end-to-end (runner + cloud persistence path) and independently re-ran the tests. No changes requested.

**What it does, verified correct**
- `extract_agent_message_text` surfaces only the per-message narrative (Claude `assistant/message` text blocks; Codex `item/completed` → `agentMessage.text`) and deliberately ignores `stream_event/*` and `item/agentMessage/delta` token streams, so mirrored volume stays bounded. Confirmed no double-emit: Codex narrative lands once at `item/completed`; the delta method isn't matched.
- Routing in `handle_bridge_event` is clean — `kind`/`summary` are only consumed in the `None` arm, observability's sanitized summary still records via `note_agent_event`, and the separate exec-command-hint block below still fires for `assistant/message`/`item/completed`.
- `push_content` flushes the compact buffer before recording, preserving ordering relative to surrounding low-signal events (covered by the new `..._forwards_content_verbatim_after_compacting_prior_low_signal` test).
- `truncate` is char-boundary safe; 4000-char cap sits well under the cloud's 64 KB `MAX_EVENT_PAYLOAD_BYTES`.

**End-to-end persistence confirmed** (no API change needed): `RunEventEndpoint` stores arbitrary `kind`(≤64 chars)/`payload` into `AgentRunEvent`, and `GET .../runs/<id>?include_events=1` returns them via `AgentRunEventSerializer` (exposes `kind`+`payload`). So `agent/message` narrative genuinely reaches the AgentRun record.

**Tests re-run locally on this branch:** 7 extractor + 1 new mirror-forwarding + 77 daemon tests all pass.

**Minor / out of scope (no action needed):**
- Run-detail caps at `[:500]` events (pre-existing); a very long run could truncate the transcript. Noted, not a regression.
- No web UI renders `agent/message` events yet — the issue only required the records to *contain* the data, which this delivers. Follow-up for a transcript view.

PR #297, opened 2026-07-02
PDASHOSS01-28 add project and local dev machine work dir info into runners/detail/<runner id>

irichard00:
## What

Surface, in the runner detail page (`runners/detail/<runnerId>`), the **project** a runner is bound to and the **dev machine** it runs on, next to the existing Pod row.

Resolves the deliverable part of PDASHOSS01-28.

## Why

The detail metadata already lists the pod, but not the project that owns it nor the dev machine hosting the runner — both are useful for identifying a runner at a glance. The data already exists server-side; the detail endpoint just didn't expose it.

## Changes

**API**
- `PodMiniSerializer`: expose `project` (FK uuid) and `project_identifier` (slug) — the same read-only `source="project.identifier"` pattern already used by `PodSerializer`.
- New `DevMachineMiniSerializer` (`id`, `host_label`, `label`) and `RunnerSerializer.dev_machine_detail`.
- `select_related("pod__project", "dev_machine")` on the runner list and detail endpoints so the new nested reads don't add N+1 queries.

**Types**
- `IPodMini.project` / `project_identifier`, new `IDevMachineMini`, `IRunner.dev_machine_detail`.

**Web**
- Render "Project" and "Dev machine" MetaRows on the detail page; `dev_machine_detail` falls back `label → host_label → —`.
- Update the detail-page and pod-filter test fixtures + assertions.

## Deferred — work dir (→ PDASHOSS01-48)

The issue also asked for the runner's *local working directory*. That value never leaves the daemon today — the path typed in the add-runner modal only builds the `--working-dir` CLI flag and lives in the daemon's local `config.toml`; enrollment/heartbeat never transmit it and `normalize_cwd()` documents that "the cloud ignores caller-selected cwd." Surfacing it requires a daemon→cloud reporting change (Rust payload + a persisted field + migration), split out to **PDASHOSS01-48** per the issue discussion.

## Validation

- `apps/web` vitest: `runner-detail-page` + `use-selected-pod-filter` pass (10 tests).
- Web `check:types` passes after rebuilding `@pi-dash/types`; changed files pass oxfmt/oxlint (pre-commit).
- Backend: change is purely additive read-only serializer fields (mirroring the shipping `PodSerializer.project_identifier`) plus `select_related`; `py_compile` clean. The full Django suite requires Postgres + the pinned stack (not provisioned in the agent sandbox) — it runs in CI on this PR.


irichard00:
**Automated review — one finding, fixed in `340a21d2`.**

The change is clean and correct: `PodMiniSerializer.project_identifier` / `DevMachineMiniSerializer` follow the existing `PodSerializer.project_identifier` pattern, `Pod.project` and `DevMachine.host_label`/`label` are all non-null on the model (so the new non-nullable `IPodMini`/`IDevMachineMini` fields are accurate), and null FKs are handled gracefully by DRF's nested-serializer traversal and the frontend `||` fallbacks. `select_related("pod__project", "dev_machine")` on both runner endpoints is correct.

**N+1 on the runs list endpoint (fixed).** `PodMiniSerializer` is also nested by `AgentRunSerializer`, whose list endpoint (`runner/views/runs.py`, `AgentRunSerializer(qs[:200], many=True)`) does not `select_related` the pod. The new `project.identifier` access therefore issued one extra project query per row (up to 200). Added `.select_related("pod__project")` to that queryset in `340a21d2`.

CI is green (pytest + web `test` pass). No other issues found.

PR #300, opened 2026-07-05
PDASHOSS01-50 Runner auto spawn instead of manually run pidash cli

irichard00:
## What & why

Automates runner creation (issue PDASHOSS01-50). Previously the "Add runner" web modal only rendered a `pidash runner add …` command the operator had to copy and run on the dev machine. This PR delivers the full cloud-driven flow: the daemon holds a machine-level control session, and the web modal can target a connected dev machine and have it create (and hot-start) the runner directly.

Originally planned as three PRs; all three phases now live here.

## Phase 1 — machine-level control session (cloud foundation)

A per-dev-machine control session — the twin of the per-runner session, keyed on `DevMachine` and authenticated by the shared `mt_` `MachineToken`. Exists even when the machine hosts **zero** runners — exactly the "add your first runner" case.

- **`MachineSession` model** + migration `0019` — one active session per machine (partial unique).
- **`services/machine_outbox.py`** — machine-keyed Redis Streams outbox (`machine_stream:{dev_machine_id}`). `create_runner` / `config_push` are offline-reject so commands aimed at an offline machine fail fast.
- **`services/pubsub.send_to_machine()`**.
- **`views/machine_sessions.py`** — open / delete / async long-poll, bound to the token's machine; poll refreshes machine presence.

## Phase 2 — `create_runner` command (cloud enqueue + daemon)

Cloud:
- `POST /api/runners/dev-machines/<mid>/create-runner/` (web session auth) — validates (project, runner name, agent) and enqueues `create_runner`; `409 machine_offline` fail-fast.
- `GET …/create-runner/<request_id>/` — status poll for the modal (short-TTL Redis result key).
- `POST /api/v1/runner/dev-machines/<mid>/commands/<request_id>/result/` (machine token auth) — daemon write-back.
- `DevMachine` list gains **`control_online`** (active machine session seen within 90s).

Daemon (Rust):
- **`daemon/machine_control.rs`** — opens the machine session with the `mt_` bearer and long-polls it: open→poll→evicted→reopen with backoff, mid dedupe, ack-by-stream-id.
- `create_runner` handler: registers via the existing X-Api-Key create endpoint, persists the `[[runner]]` block (cloud-side rollback on local failure), **hot-adds the runner in-process — no daemon restart** — and reports ok/error back.
- Supervisor's per-instance spawn extracted into `spawn_instance_tasks` + `RunnerSpawnCtx` so startup and hot-add share one code path.
- Zero-runner startup is now allowed when machine control is available.

## Phase 3 — web modal

- Connected dev machine picker (auto-selects the first connected machine; "Run `pidash runner add` manually" stays one click away).
- "Generate command" → **"Generate Runner"**.
- Machine-targeted submit drives create → status poll → created / failed / timeout panels, with a manual-command fallback on failure.

## Testing

- **Django**: 26 unit tests green (14 foundation + 12 new: enqueue online/offline/validation, result write-back + machine auth boundary, status poll, `control_online` annotation). Full `tests/unit/runner/` suite: 383 passed.
- **Rust**: 420 lib tests green incl. new wire-parsing / agent-kind / zero-runner-gate tests. Clippy clean.
- **Web**: 42 tests green incl. new remote-create success and offline-fallback tests. Types + oxlint clean.
- **Live end-to-end** against a local stack: daemon with zero runners opened the machine session (`control_online=true`), cloud pushed `create_runner`, the daemon registered + hot-spawned the runner (came up `online` with an active session, correct pod + dev machine binding), and the result key read `ok` — all without a daemon restart.


irichard00:
## Review — machine-level control session (PR 1 of 4)

Reviewed the diff against the proven per-runner session infra it mirrors. **Looks good to merge.** CI is green (pytest pass, 14 new tests), it's mergeable, the auth boundary (machine A's `mt_` token can't drive machine B) is enforced and tested, and the async long-poll handling correctly slices `BLOCK` so it never issues `BLOCK 0` on an expired deadline. The offline-reject of `create_runner`/`config_push` (fail-fast instead of silently firing minutes later) is the right call.

### Reconciling the automated (Gemini) review

I checked all four Gemini findings against the reference implementation they're modelled on, and each is a place where the new code **exactly mirrors** the existing, proven per-runner code:

| Gemini finding | Reference it mirrors | Verdict |
|---|---|---|
| `list(body.get("ack"))` splits a string / crashes on int (`machine_sessions.py:196`) | Identical to `views/sessions.py:354` — the daemon protocol always sends `ack` as a JSON list | Not a defect; matches the wire contract & convention |
| Wrap `xack` in try/except (`machine_outbox.py:322`) | `outbox.py:526` `ack_for_session` has no try/except either | Intentional mirror |
| Wrap `set` in try/except (`machine_outbox.py:230`) | `outbox.py:428` `mark_pel_drained` has no try/except | Intentional mirror |
| Wrap `exists` in try/except (`machine_outbox.py:237`) | `outbox.py:437` `is_pel_drained` has no try/except | Intentional mirror |

Applying these only to `machine_outbox.py` would make the new files *diverge* from the reference they're meant to twin, for no functional gain — the whole point of this foundational PR is to be a faithful parametrized copy so the Rust daemon (PR 2) can rely on identical semantics. Gemini's underlying "best-effort side effects shouldn't crash the hot path" principle is reasonable, but if we want it, it should land as a **separate uniform hardening pass across both the runner and machine outboxes together**, not as a one-sided divergence here. Declining for this PR; not tracking as a blocker.

No correctness issues found in my own pass. 👍

irichard00 (commented):
## Code review — APPROVE-equivalent (comment because reviewer == PR author)

**Verdict:** No CRITICAL or HIGH findings. Architecture cleanly mirrors the existing per-runner session (model/outbox/session views, HttpLoop-style daemon loop), validation is green across all three layers, and the live e2e (zero-runner daemon → machine session → `create_runner` → in-process hot-add → runner online → result `ok`) passed. The findings below are hardening follow-ups, none blocking.

### MEDIUM
1. **Result write-back not bound to the originating machine** — `views/machine_commands.py` `MachineCommandResultEndpoint`. The `mt_` token is checked against the URL's machine, but not that the `request_id` was issued *for that machine*: a token for machine B could overwrite machine A's result if it learned the uuid4 `request_id` (unguessable, so low exploitability; impact limited to forging the modal's ok/error panel). Fix: store `dev_machine_id` in the pending marker at enqueue and validate on write-back.
2. **Silent enqueue failure reads as success** — `send_to_machine` swallows non-offline Redis failures, so the endpoint returns 202 and the modal polls "pending" to a 90s timeout. Surface a 503 instead.
3. **Hot-added runner invisible to IPC/TUI until restart** — documented gap in `RunnerSpawnCtx::add_runner`; worth a follow-up issue to make the IPC instance map live like `mailboxes`/`hello_runners`.

### LOW
4. `seen_mids.clear()` at `SEEN_CAP` wipes the whole at-least-once guard; a crash-redelivery straddling the clear could double-execute `create_runner`. Consider generational/LRU eviction.
5. If the in-process hot-add fails *after* cloud registration + config write, the modal says "failed" but the runner exists and will start on the next daemon restart — consider a distinct message.
6. `machine_commands.py` imports underscore-private helpers from three sibling view modules (`_RUNNER_NAME_RE`, `_auth_dev_machine`, `_machine_is_in_workspace_scope`) — promote to a shared module.
7. Legacy `pidash connect` installs lack `dev_machine_id`/`[cli].token`, so machine control silently stays off (one info log). Add a `pidash doctor` check + docs pointer to `pidash auth login`.
8. Modal picker silently snaps back to "Run manually" if the selected machine drops offline mid-form — a hint would help.

### Validation
| Check | Result |
|---|---|
| cargo clippy / cargo test (lib) | Pass / 420 passed |
| pytest `tests/unit/runner/` | 383 passed |
| vitest (apps/web) | 42 passed |
| tsc (web/services/types) + oxlint | Pass |
| makemigrations --dry-run | PR change additive; observed drift pre-exists on main |
| Live e2e (cloud → daemon → hot-add → result) | Pass |

Full artifact: `.claude/PRPs/reviews/pr-300-review.md` on the branch owner's machine.

PR #301, opened 2026-07-11
PDASHOSS01-51 Unify Pi Dash AI and AI runner chat to use the same shared UI components

irichard00:
### Context

The primitive chat components — `ChatMessage`, `ChatMessageList`, `ChatComposer`, `ChatToolActivity` — were **already extracted and shared** between the Pi Dash AI assistant chat and the runner agent chat (they live in `apps/web/core/components/chat/` and both surfaces import them; that unification landed in #269).

What remained was the top-level `ChatContainer` the issue explicitly calls for: both full chat views still hand-assembled the outer flex-column shell (message list + optional header/footer + inline error + composer), so any layout/scroll/streaming-shell change had to be duplicated across the two files.

### Change

- Add `apps/web/core/components/chat/container.tsx` — a shared `ChatContainer` composable that assembles the message list, an optional `header` slot, an optional `listFooter`, an optional inline `error`, and the `composer` into the standard full-height chat layout.
- Route `AssistantChatRoot` and the runner chat page (`runners/chat/[runnerId]/page.tsx`) through it.
- Surface-specific differences are supplied via props/slots rather than forking the layout:
  - **Runner**: `header` (runner title/pod/status Badge + close button), `listFooter` (debug event strip), `className="min-h-[640px]"`, runner composer props.
  - **Assistant**: setup-card/empty-state, inline `error` line, assistant composer props.

A single change to the chat shell (layout, scroll behavior, error placement) now lands in one place for both surfaces.

### Scope notes

- The assistant landing page (`assistant/page.tsx`) uses only a standalone `ChatComposer` (a centered "start a chat" screen, not a full chat view) and is intentionally left as-is.
- **Follow-up candidate (out of scope):** promoting the shared chat components from `apps/web/core/components/chat/` into `packages/ui` (per the `@pi-dash/ui` convention in AGENTS.md). That is a larger, separable refactor because the components depend on app-level `MarkdownRenderer` and react-router `Link`, which would need to be injected.

### Validation

- `pnpm turbo run check:types --filter=web` — passes.
- `pnpm turbo run check:lint --filter=web` — 0 errors (pre-existing warnings unchanged).
- `oxfmt --check` on all three touched files — clean.
- No behavioral or visual change intended: the exact class strings each view used are preserved (the runner's `min-h-[640px]` overrides the container default via tailwind-merge).

### DIFF
--- a/apps/web/app/(all)/[workspaceSlug]/runners/chat/[runnerId]/page.tsx
+++ b/apps/web/app/(all)/[workspaceSlug]/runners/chat/[runnerId]/page.tsx
@@ -14,8 +14,8 @@ import { RunnerService, getRunnerDetail } from "@pi-dash/services";
 import type { IAgentChatEvent, IAgentChatMessage, IAgentChatSession, IRunner } from "@pi-dash/types";
 import { Badge, Button } from "@pi-dash/ui";
 import { ChatComposer } from "@/components/chat/composer";
+import { ChatContainer } from "@/components/chat/container";
 import { ChatMessage } from "@/components/chat/message";
-import { ChatMessageList } from "@/components/chat/message-list";
 import { useAgentChatEvents } from "@/components/runners/chat/use-agent-chat-events";
 import { useWorkspace } from "@/hooks/store/use-workspace";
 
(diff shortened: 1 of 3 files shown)


irichard00:
## Review — approved ✅

Reviewed the diff for correctness and faithfulness. This is a clean, minimal consolidation: the shared `ChatContainer` composable assembles the standard `header → message list → inline error → composer` shell, and both `AssistantChatRoot` and the runner chat page route through it via props/slots. No forking of the layout.

Verified:
- **No visual/behavioral regression.** Each view's slot content and render order match the original inline markup exactly.
- **Runner min-height preserved.** `cn` is built on `extendTailwindMerge`, so `className="min-h-[640px]"` correctly overrides the container's default `min-h-0`; the assistant view keeps `min-h-0`.
- **Types are faithful.** `ChatContainer<M extends ChatListItem>` matches `ChatMessageList`'s generic constraint and forwards every prop (`renderMessage`, `emptyState`, `listFooter` → `footer`). Consumers pass their existing message arrays unchanged.
- **Error slot equivalence.** The assistant's `{error && …}` becomes `error={error ? … : undefined}` — same output; slot renders in the same position (between list and composer).

Optional nit (non-blocking): the `error` prop is a generic `ReactNode` "inline notice" slot; the name reads slightly narrow, but the JSDoc clarifies intent, so leaving as-is is fine.

The follow-up noted in the description — promoting these components into `packages/ui` — is correctly scoped out (blocked on app-level `MarkdownRenderer` / react-router `Link` deps).

LGTM — ready to merge.

PR #309, opened 2026-07-22
PDASHOSS01-48 Report runner local dev-machine working directory to the cloud and surface it in runner detail

irichard00:
## What

Report each runner's **local dev-machine working directory** to the cloud and surface it on the runner detail page. Previously the runner reported `os` / `arch` / `version` at session-open but not the directory it actually works in.

Resolves **PDASHOSS01-48**.

## How

Follows the existing session-open reporting path end-to-end:

**Runner (Rust)**
- Add `working_dir` to `AttachBody` (the session-open body).
- Populate it in `attach_body_for_instance` via a new `resolve_working_dir` helper:
  - **Pooled** runners (those referencing a `[[workdir]]`) report the workdir's **canonical clone path** (`PoolHandle::canonical()`).
  - **Legacy** runners report `workspace.working_dir`.
- Normalize reported paths to absolute paths before sending them.
- If a pooled runner has no pool, report no working directory instead of its vestigial legacy path.
- The retired WS `ClientMsg::Hello` path is left untouched — only the HTTP session-open body is live.

**Cloud (Django API)**
- Add an extensible `dev_metadata` `JSONField` to the `Runner` model (+ migration `0020`).
- Whitelist and persist the reported path as `dev_metadata["working_dir"]` in `session_service.apply_hello`, preserving existing development metadata and limiting the path to 1024 characters. Omitted fields from older runners preserve the last value; an explicit empty value clears stale path metadata.
- Expose `dev_metadata` via `RunnerSerializer`.

**Web**
- Add typed `IRunnerDevMetadata` and `IRunner.dev_metadata` to the runner API contract.
- Read `runner.dev_metadata.working_dir` in the **"Working directory"** row on the runner detail metadata panel.

## Testing

- `cargo test resolve_working_dir` — 2 regression tests pass.
- `cargo check --tests` — pass.
- Changed Rust files pass `rustfmt --check`.
- Web `runner-detail-page` + `use-selected-pod-filter` Vitest: 10 tests pass.
- `@pi-dash/types` type-check passes; changed frontend files pass Oxfmt and Oxlint.
- Development-metadata merge tests cover key preservation, explicit clearing, malformed JSON recovery, and path-length limits.
- Changed Python files pass Ruff lint and Python compilation.
- Added a session-open assertion in `test_cli_auth_flow.py` that the path is persisted under `dev_metadata`. The Django test suite could not start in the authoring checkout because its Python runtime is missing Celery; CI will run it with the complete service dependencies.

## Notes

- Merged the latest `main` into this branch before publishing.
- Broader web type-check and repository-wide formatting still report unrelated pre-existing failures outside this PR's changes.


irichard00 (commented):
**Automated code review — approved ✅**

Reviewed the full diff for PDASHOSS01-48. The change reports each runner's local dev-machine working directory to the cloud at session-open and surfaces it in runner detail, cleanly mirroring the existing `os`/`arch`/`version` path end-to-end.

Verified:
- **Migration chain** — `0020` depends on `0019_machine_session`, which is the latest migration. Correct.
- **Rust** — `resolve_working_dir` looks up pools by `config.workdir` name; the `pools` map is keyed by `wd.name` (supervisor.rs:120), matching the existing `with_pool` lookup. `PoolHandle::canonical()`, `config.workdir`, and `workspace.working_dir` all exist. Legacy fallback keeps the field non-empty. `AttachBody` key-set regression test updated.
- **API** — `apply_hello` persists with `body.get(..)` fallback + `[:1024]` truncation → backward-compatible for older runners that don't send the field. Serializer exposes it read-only.
- **Web** — `IRunner.working_dir` added; "Working directory" MetaRow renders with mono/`break-all`, `—` fallback.
- **CI** — all checks green: cargo test + clippy, pytest (incl. the new session-open persistence assertion), and web tests.

No correctness, backward-compatibility, or reuse issues found. Ready for human merge.

PR #310, opened 2026-07-22
PDASHOSS01-61 Auto-suggest user timezone from browser in project creation

irichard00:
## What

New projects now default their **Project Timezone** to the user's browser-detected local timezone instead of always defaulting to UTC.

## Why

`PDASHOSS01-61` — when creating a project the timezone was always UTC (the create form never sent a `timezone`, so the backend fell back to the workspace/UTC default). Users then had to manually fix it in project settings.

## How

- `apps/web/ce/components/projects/create/utils.ts`: `getProjectFormValues()` now sets `timezone` from `Intl.DateTimeFormat().resolvedOptions().timeZone`, wrapped in a small `getBrowserTimezone()` helper that falls back to `"UTC"` if detection is unavailable or throws.
- Modern browsers canonicalize the returned zone id (e.g. `Asia/Calcutta` → `Asia/Kolkata`), so the value aligns with the backend `Project.timezone` choices (`pytz.common_timezones`). The submitted value is still user-editable in project settings.
- Mirrors the existing pattern already used in `scheduler-bindings/install-binding-modal.tsx`.

## Notes

- The CE create form does not render a visible timezone field (that field lives in project *settings*); this sets the default value that gets submitted at creation, satisfying "auto-suggest the local timezone as the default." No new UI field was added.

## Validation

- `oxlint` + `oxfmt` on the changed file: clean.
- `npm run check:types`: no new errors from this change (the only failures are pre-existing, in `tests/runners/*`, unrelated to this file).

### DIFF
--- a/apps/web/ce/components/projects/create/utils.ts
+++ b/apps/web/ce/components/projects/create/utils.ts
@@ -8,8 +8,23 @@ import { RANDOM_EMOJI_CODES } from "@pi-dash/constants";
 import type { IProject } from "@pi-dash/types";
 import { getRandomCoverImage } from "@/helpers/cover-image.helper";
 
+// Detect the user's local IANA timezone from the browser so new projects
+// default to it instead of always defaulting to UTC. Modern browsers
+// canonicalize the value returned by `resolvedOptions().timeZone`
+// (e.g. "Asia/Calcutta" -> "Asia/Kolkata"), so it lines up with the backend's
+// `pytz.common_timezones` choices. If detection is unavailable or throws, fall
+// back to UTC.
+const getBrowserTimezone = (): string => {
+  try {
+    return Intl.DateTimeFormat().resolvedOptions().timeZone || "UTC";
+  } catch {
+    return "UTC";
+  }
+};
+
 export const getProjectFormValues = (): Partial<IProject> => ({
   cover_image_url: getRandomCoverImage(),
+  timezone: getBrowserTimezone(),
   description: "",
   logo_props: {
     in_use: "emoji",


irichard00:
## Automated review — approved ✅

Traced the change end-to-end and it's correct, minimal, and matches the issue's acceptance criteria.

**Change:** `apps/web/ce/components/projects/create/utils.ts` adds `getBrowserTimezone()` and sets `timezone` in `getProjectFormValues()`.

**Verified:**
- `getBrowserTimezone()` returns `Intl.DateTimeFormat().resolvedOptions().timeZone || "UTC"` inside a `try/catch` → falls back to UTC on empty result or throw, exactly matching "if browser timezone parse fails, still use UTC."
- `timezone` is a declared field on `IProject` (`packages/types/src/project/projects.ts:61`).
- Data flow reaches the backend: `getProjectFormValues()` feeds `useForm` `defaultValues` in `create/root.tsx`, and `onSubmit` passes `formData` straight to `createProject(...)` without stripping `timezone`. Confirmed unregistered defaults survive submission via the sibling `cover_image_url` default, which `onSubmit` reads back from `formData`.
- Mirrors the existing repo pattern (`scheduler-bindings/install-binding-modal.tsx`).

**Residual risk (non-blocking):** a very old browser returning a non-canonical IANA id (e.g. `Asia/Calcutta`) that isn't in the backend's `pytz.common_timezones` set would be rejected at create time. Modern browsers canonicalize `resolvedOptions().timeZone`, so this is low and is the same risk the repo already accepts in `scheduler-bindings`. No change requested.

Ready for human review/merge.

PR #311, opened 2026-07-22
PDASHOSS01-41 Requester onboarding UX: join a workspace by admin email

irichard00:
## What & why

Issue **PDASHOSS01-41** — a new, uninvited user only ever saw "Create your workspace" with no way to join an existing one. This is **PR 2 of 3** (the requester onboarding UX), built on the merged backend (**#279**, **#303**).

A signed-in but workspace-less user can now, during onboarding, **join an existing workspace by typing a workspace admin's email** instead of creating one, then wait for approval.

## Flow

1. In the "Create workspace" onboarding step there's a new **"Join an existing workspace by admin email"** option.
2. The user types the admin's email and sends a request → `POST /api/users/me/workspaces/join-requests/`.
3. They land on a **"Waiting for approval"** holding view.
4. Because a request-but-not-yet-approved user has **no workspace**, onboarding never completes — so on every login/refresh they **bounce back to this pending view**. This is the intended "bounce back to onboarding" behaviour agreed on the issue, and it falls out naturally from the existing onboarding gate (`finishOnboarding` requires `workspacesList.length > 0`).
5. On approval (admin side — PR 3), they become a member and next login routes them into the workspace normally.

## Key decisions (all confirmed on the issue)

- **Escape hatch** — "Create your own workspace instead" on the pending view, so a slow / absent / mistyped admin never strands the user. (We deliberately don't reveal whether the typed email is a real admin, so a typo would otherwise wait forever.)
- **Already-a-member** — the backend returns `200` + a workspace slug when the email resolves only to a workspace the user is already in; the UI consumes that to route them straight in (mirrors the existing invitations flow) instead of showing pending.
- **Enumeration-neutral** — the pending screen is identical whether or not the email resolved to a real admin; the requester's own join-request list deliberately omits the workspace name/slug (backend design).

## Implementation notes

- The pending "step" is implemented as a **view inside the existing `WORKSPACE_CREATE_OR_JOIN` onboarding step** (alongside create / join-invites), so the onboarding progress machine and header are untouched — lowest blast radius on the app's critical first-run path.
- `joinRequests` are fetched on the onboarding page via SWR (same shape as invitations) and threaded down; a pending request selects the pending view on mount.
- New service methods `createWorkspaceJoinRequest` / `userWorkspaceJoinRequests`; new types `IWorkspaceJoinRequest`, `EWorkspaceJoinRequestStatus`, and two new `ECreateOrJoinWorkspaceViews` members; en translation strings registered.

## Known follow-ups (out of scope here)

- **PR 3** — admin approve/deny UI under Members → Requests + notification.
- The escape hatch does not delete the outstanding pending request (there's no delete endpoint); a stale pending request is harmless — the admin could still approve it later.

## Verification

- `turbo run check:types --filter=web` — clean for all changed files (the one remaining error, `tests/runners/runner-runs-page.test.tsx`, is pre-existing and unrelated).
- `oxlint` + `oxfmt` clean on all changed files (also enforced by the pre-commit hook).
- No backend changes; consumes the already-merged endpoints only.


irichard00 (commented):
**Review — PR 2 (requester onboarding UX). No blocking defects; ready to merge.**

I traced this against the merged backend (#279/#303) and the existing onboarding/invitations patterns. The three confirmed product decisions are all implemented correctly: the "join by admin email" entry, the pending-approval holding view, the "create your own workspace instead" escape hatch, and consuming the backend's already-member `200 + slug` to route straight in. Types are sound (`IUserLite` imported), and it follows established conventions (service methods, header component, view switch inside `WORKSPACE_CREATE_OR_JOIN` so the step machine is untouched).

I did **not** change any code. A few non-blocking notes for your consideration:

1. **Escape-hatch re-selection (low).** `WorkspaceSetupStep`'s init `useEffect` depends on `[invitations, joinRequests]` and re-selects the pending view whenever a `PENDING` request is present. In practice this is safe because SWR 2.2.4's default `dequal` comparison keeps the `joinRequests` reference stable across focus revalidations when the data hasn't changed — so a user who clicks "Create your own workspace instead" won't get yanked back to pending. Worth being aware that the guarantee rests on SWR's dedup, not on the effect deps as the inline comment implies. If the admin *denies* the request while the user sits on the pending view, the data genuinely changes and they'll silently jump to the create view with no "request denied" messaging — a small UX gap to fold into PR 3's denial handling.

2. **Already-member routing ignores the returned slug (low).** On the `200 + workspace_slug` branch, the UI calls `fetchWorkspaces()` + `fetchCurrentUserSettings()` then `finishOnboarding()`, discarding `workspace_slug` and relying on the generic post-onboarding redirect. That mirrors the invitations-accept flow, and the backend already-member branch does not set `Profile.last_workspace_id`. Worth a quick manual confirm that a member lands in the right workspace here — though this path (already a member yet still in onboarding) is itself an edge.

3. **Brief create-view flash (cosmetic).** The onboarding page gates rendering on `invitationsLoader` but not the join-requests loader, so a pending user can momentarily see the create view before SWR resolves and switches to pending. Matches the existing invitations behavior.

4. **Two join buttons (nit).** In the normal create footer, "Join existing workspace" (when invited) and "Join an existing workspace by admin email" can both appear. Functional; consider consolidating the labels.

Nice touch: the `isWorkspaceCreationDisabled` branch now offers a real join path where previously such users were fully stuck.

PR #313, opened 2026-07-22
PDASHOSS01-65 Fix issue run queue stuck after project change

irichard00:
## Problem

When an issue that already has an outstanding agent run is moved to a new project, the run gets stuck in queue forever and the issue can never be run again — matching the report in PDASHOSS01-65.

## Root cause

`move_work_item_to_project` (`apps/api/pi_dash/utils/issue_move.py`) updates `issue.assigned_pod` to the target project's default pod, but leaves the issue's existing **QUEUED** `AgentRun` rows pinned to the *source* project's pod. From there:

1. The matcher only drains a run whose `pod` matches an idle runner's pod — `matcher.next_for_runner` filters `pod=runner.pod`. A queued run left on the old pod is never picked up by the new project's runners → stuck in queue.
2. A QUEUED run still counts as the issue's single active run (`orchestration._active_run_for`), so a subsequent state transition / Run-AI **skips** creating a replacement run (`active-run-exists`) → the issue is permanently unrunnable.

(Note: `IssueCreateSerializer` already blocks pod reassignment while a run is active, but the *move* path bypasses that serializer, which is how the inconsistent state arises.)

## Fix

During the move, repoint the issue's outstanding **QUEUED** runs onto the target project's default pod and clear any `pinned_runner` (a runner pinned in the old pod is not eligible in the new one), then kick a drain of the target pod on commit so an idle runner picks the run up immediately.

Runs already `ASSIGNED`/`RUNNING` on a live runner are intentionally left in place — they are mid-flight and complete (or fail) on the original runner. Terminal runs are historical and untouched.

## Tests

New `apps/api/pi_dash/tests/unit/runner/test_issue_move_run_repoint.py`:

- `test_move_repoints_queued_run_to_target_pod` — queued run repointed, `pinned_runner` cleared.
- `test_repointed_run_is_drainable_by_target_runner` — the post-move drain assigns the run to an idle runner in the new project (end-to-end proof the stuck-in-queue symptom is gone).
- `test_move_leaves_running_run_in_place` — an in-flight run is not disturbed.
- `test_move_leaves_terminal_run_in_place` — completed runs are not repointed.

All 4 pass against a real Postgres; neighboring suites (`test_issue_pod`, `test_drain_pod`, `test_matcher`) stay green (50 passed). `ruff check`/`ruff format` clean.


irichard00:
## Review — PDASHOSS01-65 (repoint queued runs on project move)

Reviewed the diff against `main`. The fix is correct, minimal, and well-scoped. Verified each claim against the code:

- **Root cause confirmed.** `matcher.next_for_runner` filters `pod=runner.pod` (+ pinned_runner), so a QUEUED run left on the source project's pod is invisible to the new project's runners. And `orchestration._active_run_for` counts QUEUED as the issue's single active run, so no replacement run is created — the issue becomes permanently unrunnable. Both halves of the described symptom are real.
- **Fix is right.** Repointing QUEUED runs to `target_pod`, clearing `pinned_runner` (a pin into the old pod is ineligible in the new one), and kicking `drain_pod_by_id` on commit all line up with the existing matcher/drain contracts. Guarding on `target_pod is not None` matches `default_for_project_id`'s nullable return.
- **Scope is defensible.** Leaving ASSIGNED/RUNNING (BUSY) and terminal runs in place is correct — they're mid-flight on a live runner or historical.
- **Tests are solid.** Four regression tests, including an end-to-end one proving the repointed run is drained/ASSIGNED to a new-project runner.

### Non-blocking observation (optional follow-up)

The repoint targets `status=QUEUED` only. Runs in the *non-busy-but-active* states — `PAUSED_AWAITING_INPUT`, `AWAITING_APPROVAL`, `AWAITING_REAUTH` — are left pinned to the old pod. In the normal path they resume on their existing runner, so leaving them is fine. But if such a run is later hit by `resume_unavailable` / `assign_rejected_busy`, `apply_run_resume_unavailable` (run_lifecycle.py) re-queues it onto `run.pod_id` — i.e. the **old** pod — and drains that pod, reproducing the exact stuck-in-queue condition this PR fixes. It's a narrow scenario (issue moved while a run was paused/awaiting, then that session becomes unavailable), and expanding the repoint blindly would be wrong (a paused run's runner in the old pod may still hold the session). Flagging only so it's a conscious choice — a follow-up could re-point on the resume-unavailable path instead, or note the limitation.

Approving in spirit; no changes required.

PR #316, opened 2026-07-23
PDASHOSS01-66 Debug missing description field in Pi Dash CLI issue creation

irichard00:
## Problem

When an AI agent (or anyone) creates an issue with the Pi Dash CLI, the issue came out with a **title but no description**, even when `--description` was supplied.

## Root cause

Neither of the two hypotheses in the issue was quite right — the CLI *does* expose `--description` and the docs *do* explain it. The bug was the JSON field name on the wire.

`pidash issue create` / `issue patch` sent the text under a bare **`description`** key:

```rust
body.insert("description".into(), Value::String(desc));
```

The web API (`apps/api/pi_dash/app/views/issue/base.py`) feeds the request body straight into `IssueCreateSerializer`, a DRF `ModelSerializer` over the `Issue` model. That model's only description column is **`description_html`** (plus the derived `description_json` / `description_stripped` / `description_binary`) — there is **no `description` field**. DRF silently ignores unknown write keys, so the description was dropped and the issue saved with the default empty body (`<p></p>`).

The web editor, the in-app AI assistant tool, and the GitHub sync importer all write `description_html`; the CLI was the only client sending the wrong key. The PATCH view even keys its description-version bookkeeping off `request.data.get("description_html")`, confirming `description_html` is the intended contract.

## Fix

Convert the plain-text/markdown `--description` input to minimal HTML and send it as `description_html` in both `cmd_create` and `cmd_patch`. The converter mirrors the server-side renderer (`assistant/runtime/markdown.py::markdown_to_html`):

- blank line → new `<p>` paragraph
- single newline → `<br/>`
- HTML-escaped (`& < > " '`)
- empty input → `<p></p>` (the model default, so `--description ""` clears the body)

The server re-sanitizes `description_html` via `validate_html_content` and the `Issue` model re-derives `description_stripped` on save, so the CLI only needs to produce correct HTML, not defensive HTML. No change to the shared web serializer, so existing web create/update flows are untouched.

## Tests

- 7 new unit tests for `description_to_html` (single/multi paragraph, `<br/>`, HTML escaping, empty, trimming).
- `cargo test --lib cli::issue`, `--test pidash_cli_contract`, `--test pidash_cli_structure` all pass. (One unrelated pre-existing `workspace::git` test fails locally due to a macOS `/private/var` vs `/var` symlink path quirk.)

### DIFF
--- a/runner/src/cli/issue.rs
+++ b/runner/src/cli/issue.rs
@@ -229,7 +229,16 @@ async fn cmd_create(
     let mut body: Map<String, Value> = Map::new();
     body.insert("name".into(), Value::String(args.title));
     if let Some(desc) = args.description {
-        body.insert("description".into(), Value::String(desc));
+        // Issue descriptions are stored as rich text: the API's serializer is a
+        // ModelSerializer over the `Issue` model, whose only description column
+        // is `description_html` (it has no plain `description` field). Sending a
+        // bare `description` key was silently dropped, so CLI-created issues had
+        // an empty body. Convert the plain-text/markdown input to minimal HTML
+        // and send it under the key the server actually persists.
+        body.insert(
+            "description_html".into(),
+            Value::String(description_to_html(&desc)),
+        );
     }
     if let Some(prio) = args.priority {
         body.insert("priority".into(), Value::String(prio));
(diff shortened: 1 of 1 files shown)


irichard00:
**Review: APPROVED** ✅ (automated post-merge review, PDASHOSS01-66)

Verified the diagnosis and fix against the current server code — all four load-bearing claims hold:

1. **Issue model has no plain `description` column** — `apps/api/pi_dash/db/models/issue.py:137-140` defines `description_json` / `description_html` / `description_stripped` / `description_binary` only. ✔
2. **Bare `description` is silently dropped** — `IssueCreateSerializer` (`apps/api/pi_dash/app/serializers/issue.py:120`) is a `ModelSerializer` over `Issue` using `exclude=["workpad","assigned_pod"]`. Exclude-based field generation means `description_html` is writable and an unknown `description` key is ignored by DRF. ✔
3. **`description_html` is the real contract** — `IssueViewSet` uses this serializer for create/update (`app/views/issue/base.py:208`), and the PATCH path gates description-version bookkeeping on `request.data.get("description_html")` (`:625`). ✔
4. **Converter parity** — `description_to_html` matches the server renderer `assistant/runtime/markdown.py::markdown_to_html` exactly: same `\n\n` paragraph split, `\n`→`<br/>`, and HTML escaping equivalent to Python's `html.escape(quote=True)` (`'`→`&#x27;`, `"`→`&quot;`, plus `& < >`). ✔

**No defects found.** The 7 converter unit tests cover the meaningful cases (single/multi-paragraph, `<br/>`, escaping, empty-clears, trim/collapse).

Two non-blocking notes for the record:
- The CLI sends only `description_html` and leaves `description_json` empty. This is **not** a regression — the in-app AI assistant and the GitHub-sync importer do the same (see the `markdown.py` module docstring: no Python Tiptap converter exists). The CLI now behaves like those existing non-web writers.
- CRLF (`\r\n\r\n`) blank lines don't split into paragraphs in either the CLI converter or the server renderer — behavior is identical on both sides, so no divergence is introduced.

PR #320, opened 2026-07-28
PDASHOSS01-69 Fix PR state sync from GitHub to pidash

irichard00:
## Problem

PR links on Pi Dash work items show a stale state — e.g. a PR shows **open** in Pi Dash while it is actually **merged** on GitHub.

## Root cause

PR display state is refreshed **only** by the GitHub App `pull_request` webhook (`_refresh_pr_links`); there is no reconciliation/polling fallback. The webhook endpoint (`GithubAppWebhookEndpoint.post`) recorded the `GithubWebhookDelivery` row with status `RECEIVED` **before** processing, and then short-circuited **every** duplicate `X-GitHub-Delivery` with a 202 — regardless of whether the prior attempt actually applied the update:

```python
if not created:
    return Response({"status": delivery.status}, status=202)
```

GitHub reuses the same delivery GUID for automatic retries (on timeout / 5xx) and for a manual **Redeliver**. So a delivery whose first attempt was:
- **interrupted** mid-request (deploy / pod restart → row stuck `RECEIVED`), or
- **threw** (caught → `FAILED`),

was never re-applied — the redelivery was swallowed with 202. And because the endpoint always returns 202, GitHub never sees an error to retry against either. A single dropped merge (`closed` + `merged`) event therefore leaves the PR stuck showing **open** indefinitely, matching the reported symptom.

The state-mapping logic itself (`pr_snapshot_from_payload` → `_refresh_pr_links`) is correct; the bug is purely in idempotency handling.

## Fix

Only fast-return when the prior attempt reached a **terminal-success** outcome (`PROCESSED` / `SKIPPED`); **reprocess** `RECEIVED` and `FAILED` deliveries on redelivery. Reprocessing is safe: `_refresh_pr_links` is idempotent and guarded against out-of-order payloads by the PR's `updated_at`. A successful reprocess also clears any stale `error` from the prior failed attempt. This preserves idempotency for GitHub's normal at-least-once duplicate deliveries while letting a missed update recover.

## Tests

Added contract tests in `test_github_pr_webhook.py`:
- redelivery of a `RECEIVED`/`FAILED` prior delivery reprocesses and applies the snapshot (and clears stale error);
- redelivery of a `PROCESSED`/`SKIPPED` prior delivery fast-returns and leaves the snapshot untouched.

Full run: **26 passed** (`test_github_pr_webhook.py`, `test_github_pr_link_app.py`, `test_github_pr_link.py`).

## Follow-up (not in this PR — needs a design decision)

There is still no reconciliation path, so drift cannot self-heal if a webhook is **never delivered** (app not subscribed to `pull_request`, event dropped entirely, or the app was down when the event fired). Recommended follow-up: a periodic Celery refresh or an on-render best-effort fetch, reusing existing building blocks (`best_effort_snapshot`, `GithubClient.get_pull_request`, adapter `get_code_review`). Per the issue, this ticket is intentionally kept open.

### DIFF
--- a/apps/api/pi_dash/app/views/integration/github.py
+++ b/apps/api/pi_dash/app/views/integration/github.py
@@ -890,10 +890,23 @@ class GithubAppWebhookEndpoint(BaseAPIView):
                 "status": GithubWebhookDelivery.Status.RECEIVED,
             },
         )
-        if not created:
+        # GitHub reuses the same X-GitHub-Delivery GUID for automatic retries
+        # (on timeout / 5xx) and for a manual "Redeliver". Only fast-return when a
+        # prior attempt reached a terminal-success outcome; a delivery left in
+        # RECEIVED (the request was interrupted mid-processing — e.g. a deploy or
+        # pod restart) or FAILED (an exception was caught below) must be
+        # reprocessed so the snapshot can still converge. Reprocessing is safe:
+        # `_refresh_pr_links` is idempotent and guarded against out-of-order
+        # payloads by the PR's `updated_at`. Without this, a single dropped
+        # `pull_request` (merge) delivery leaves the PR stuck showing "open".
+        if not created and delivery.status in {
+            GithubWebhookDelivery.Status.PROCESSED,
+            GithubWebhookDelivery.Status.SKIPPED,
+        }:
             return Response({"status": delivery.status}, status=status.HTTP_202_ACCEPTED)
 
         try:
+            delivery.error = ""
             if event == "ping":
                 delivery.status = GithubWebhookDelivery.Status.PROCESSED
             elif event == "pull_request":
@@ -922,7 +935,7 @@ class GithubAppWebhookEndpoint(BaseAPIView):
             else:
                 delivery.status = GithubWebhookDelivery.Status.SKIPPED
             delivery.processed_at = timezone.now()
-            delivery.save(update_fields=["status", "processed_at", "updated_at"])
+            delivery.save(update_fields=["status", "error", "processed_at", "updated_at"])
         except Exception as e:
             log_exception(e)
             delivery.status = GithubWebhookDelivery.Status.FAILED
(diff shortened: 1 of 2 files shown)


irichard00 (commented):
**Automated review — PDASHOSS01-69**

I re-read the fix against `_refresh_pr_links` and the delivery model. The diagnosis and the fix hold up:

- **Root cause is right.** `if not created: return 202` swallowed every duplicate `X-GitHub-Delivery`, including redeliveries of a delivery that never terminally applied (`RECEIVED` after a mid-request interrupt, or `FAILED`). Since GitHub reuses the GUID for auto-retries and manual "Redeliver", a dropped merge event never reconverged → PR stuck showing "open".
- **Fix is safe.** Gating the fast-return on terminal-success (`PROCESSED`/`SKIPPED`) and reprocessing `RECEIVED`/`FAILED` is sound because `_refresh_pr_links` is idempotent and guarded against out-of-order payloads by `pr_updated_at`/`remote_updated_at`. Resetting `delivery.error = ""` and adding it to `update_fields` correctly clears stale errors on a successful reprocess.
- **Tests** cover both directions (non-terminal redelivery reprocesses; terminal redelivery is untouched) and assert the snapshot outcome, which is the behavior that matters.

**Nit (non-blocking, no change requested):** the endpoint has no row lock, so if GitHub auto-retries the same GUID while the first request is still in-flight (row still `RECEIVED`), both attempts now run `_refresh_pr_links` concurrently where before the second fast-returned. This is safe given idempotency + the `updated_at` guard — just flagging it as a known, acceptable trade-off.

**Scope note (already flagged on the issue):** this makes a *delivered* webhook recoverable, but drift still can't self-heal if a `pull_request` webhook is never delivered (App unsubscribed / down at merge time). The reconciliation follow-up called out in the issue thread is the remaining gap.

LGTM to merge.

PR #321, opened 2026-07-28
PDASHOSS01-68 Do not move issues from "in progress" or "in review" to "done"

irichard00:
## Problem

When a Pi Dash issue is In Progress we auto-trigger a runner AI agent to work it. For question / debug / investigation tasks ("what color is the home page button?", "help debug why X"), the agent answers correctly in a comment but then also flips the issue to **Done** — because from its view the task looks finished. Once an issue is Done it drops off the user's radar. That promotion should belong to a human (or a separate supporting process), not the runner.

Per the issue discussion (Comments 2 & 3): the runner — **the AI agent via its prompt _and_ the backend ticking code** — must never proactively move an issue from In Progress or In Review to Done. Additionally, reduce the In Review ticking budget from 8 to 4, and when review ticking is exhausted leave the issue **In Review** for a human to close.

## Changes

**Prompt sections** (`apps/api/pi_dash/prompting/sections/` — the live composed prompt via `composer.build_first_turn`):
- **coding-task**: a finished issue moves to **In Review** whether or not it opened a PR. The runner never emits `--state "Done"`. Reworded `ending-run.md`, `default-posture.md`, `implementation.md`, `pidash-cli.md`, calling out the classic question/debug examples explicitly.
- **review**: an **approved** review posts its summary and **leaves the issue In Review** — never moves it to Done (`review-cycle.md` Step 3, `ending-run.md` review branch, `pidash-cli.md`).

**Backend**:
- Reduce the default In Review ticking budget **8 → 4** (`Project.agent_review_default_max_ticks`, migration `0155`). Per-issue overrides untouched; existing projects keep their stored value.
- Exclude the In Review phase from the cap-hit **auto-pause** in `scheduling.maybe_apply_deferred_pause`: when the review budget is exhausted the issue stays In Review for a human instead of being reparked to Paused. (Verified there is no code path that moves an issue to Done on review approval — `done_signal`'s `requested_group` is stored but never acted on; the "by code itself" concern reduces to this auto-pause.)

## Tests

- `test_composer`: both coding-task and review prompts assert `--state "Done"` is absent and route to In Review.
- `test_scheduling`: new `test_deferred_pause_skips_when_issue_is_in_review`.
- `test_issue_agent_ticker`: new `test_review_max_ticks_schema_default_is_four`.
- All 63 targeted tests pass against a freshly-migrated test DB (validates migration `0155`).


irichard00:
**Review pass (automated) — approve.**

Verified PR #321 against the four requirements from the issue thread (Comments 2 & 3):

1. **Runner never proactively moves an issue to Done (prompt).** The `--state "Done"` example is gone from every reachable coding-task and review path (`ending-run.md`, `default-posture.md`, `implementation.md`, `pidash-cli.md`); the only remaining "Done" mentions are prose forbidding it. In Review is now the sole success routing for both PR and non-code (question/debug/investigation) finishes. ✅
2. **Review cycle no longer auto-completes to Done — prompt *and* code.** `review-cycle.md` Step 3 approved-outcome leaves the issue In Review. I traced the code side: `done_signal.py` stores `requested_group` but nothing consumes it (`grep` for `requested_group` outside its definition = 0 hits), so there was no code path completing an approved review to Done. The only code that *repositioned* a review issue at cap-hit was the auto-pause, now gated below. ✅
3. **Review ticking 8 → 4.** `Project.agent_review_default_max_ticks` default + migration `0155`. ✅ (see note)
4. **Budget exhaustion leaves the issue In Review, not Paused.** The new `StateGroup.REVIEW` early-return in `maybe_apply_deferred_pause` is correctly placed (In Progress still auto-pauses) and reachable — `agent_phases.PHASES` registers "In Review" as a ticking state, so cap-hit on the review phase really would have hit the auto-pause before this guard. ✅

Tests are consistent with the changes (both prompt kinds assert Done is never emitted; new review-phase pause-skip test; 4-tick schema-default test). I could not execute the suite in this review environment (missing app deps / no venv), so I relied on static review + the implementation run's reported 63 passing.

**One non-blocking note (product judgment, not a code defect):** migration `0155` is a bare `AlterField(default=4)`, which only affects **newly created** projects. Existing projects already materialized `8` into their row and keep it — including projects that never customized the value. So on any existing deployment the 8→4 reduction won't take effect until a project is recreated. The migration docstring documents this as intentional (avoid clobbering an operator's stored value), and there's no clean way to distinguish "operator chose 8" from "8 was the old default." Flagging so the human can decide whether a one-time data migration for rows still at `8` is wanted; leaving as-is is defensible.

Minor: `seed.py`'s `REVIEW_TEMPLATE_BODY` still references the legacy `pi-dash-done` done-signal, but it's off the live rendering path (the prompt composes from `sections/`), so it's harmless — cleanup candidate only.

Leaving the issue **In Review** per the new policy this PR itself establishes.

PR #322, opened 2026-07-30
PDASHOSS01-70 Change review state tick cadence to every 8 hours

irichard00:
## What

Change the default **In Review** ticking cadence from every 3 h to every 8 h.

- `Project.agent_review_default_interval_seconds`: `10800` (3 h) → `28800` (8 h)
- New migration `0156_review_interval_8h.py` alters the field to match

## Why

For the In Review state the runner should re-review less often, leaving more room for a human to weigh in between passes.

## Notes

- **No data backfill** — mirrors the sibling change `0155_reduce_review_max_ticks.py` (PDASHOSS01-68): projects created before this change already materialized `10800` into their row, and silently rewriting an operator's stored value is more surprising than leaving it. New projects get `28800`.
- In Progress cadence (`agent_default_interval_seconds`) is unchanged.
- Fixed the now-stale `agent_review_default_max_ticks` comment (4 ticks × 8 h = 32 h window).

## Validation

This runner has no project Python venv (system Python lacks `corsheaders`), so `makemigrations --check`/pytest can't boot Django settings here. Validated statically: byte-compile OK; model default == migration default (28800); clean linear migration leaf (0155 → 0156); existing ticker tests set the value explicitly so they're unaffected.

### DIFF
--- a/apps/api/pi_dash/db/migrations/0156_review_interval_8h.py
+++ b/apps/api/pi_dash/db/migrations/0156_review_interval_8h.py
@@ -1,0 +1,40 @@
+# Copyright (c) 2023-present Pi Dash Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
+"""Lengthen the default In Review ticking interval from 3 h to 8 h.
+
+Per PDASHOSS01-70 the review-phase cadence is stretched so the runner
+re-reviews an In Review issue every 8 h instead of every 3 h. Each review
+tick is a full re-read of the thread and PR; a wider gap leaves more room
+for a human to weigh in between passes without exhausting the review
+budget prematurely.
+
+Only the project-level default changes; per-issue
+``IssueAgentTicker.review_interval_seconds`` overrides are untouched, and
+the In Progress cadence (``agent_default_interval_seconds`` = 10800) is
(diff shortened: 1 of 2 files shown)


irichard00 (commented):
**Review: approved** ✅ (automated review pass)

Focused, correct, and mirrors the sibling `0155` precedent from PDASHOSS01-68.

Verified:
- `Project.agent_review_default_interval_seconds` default `10800`→`28800` (8 h). This is the field consumed by the In Review branch of `IssueAgentTicker.effective_interval_seconds()`, so new projects' review phase now ticks every 8 h.
- Migration `0156_review_interval_8h` is an `AlterField` to `28800`, depends on `0155` and is the clean linear leaf; model default == migration default.
- Stale `agent_review_default_max_ticks` window comment corrected `12 h`→`32 h` (4 × 8 h ✓).
- In Progress cadence (`agent_default_interval_seconds = 10800`) untouched.
- No data backfill — matches `0155`'s stated rationale (don't silently rewrite operators' stored values).
- Existing ticker tests set the review interval explicitly in the fixture, so none regress.

Non-blocking suggestion (optional): `0155` paired its change with `test_review_max_ticks_schema_default_is_four`. For parity, consider adding a `test_review_interval_schema_default_is_28800` asserting a freshly-created project inherits `28800`, to guard the new default against regression. Not required for merge.

PR #323, opened 2026-07-31
PDASHOSS01-71 Add labels to prompt sections and receipts

irichard00:
## What

Adds workspace-scoped, user-managed **labels** for prompt **sections** and **receipts** (issue PDASHOSS01-71). A label can be attached to many sections/receipts and a section/receipt can carry many labels (n:n). Labels render as clickable chips; clicking one filters the prompts page down to everything carrying it.

## Design decisions (defaults — see issue thread)

The issue had no acceptance criteria and the design clarification went unanswered across several ticks (runner auth failures), so this proceeds with the documented, reversible defaults:

- **Workspace-scoped, shared** labels (mirrors `PromptSectionOverride` scoping).
- **Admin-managed** writes (create/edit/delete + attach/detach); reads for any member — mirrors how section overrides are admin-gated.
- A **dedicated `PromptLabel` model**, kept separate from the issue `Label` (which is project/issue-coupled and hierarchical).
- Assignments key off the **code-defined `section_key` / receipt `kind`** strings, because sections (`prompting/registry.py`) and receipts (`recipes.py`) are not DB rows.
- Clicking a chip **filters the current prompts tab in place**.

## Backend (`apps/api/pi_dash/prompting/`)

- `PromptLabel` + `PromptLabelAssignment` models + migration `0005_prompt_label.py`.
- REST under `/api/workspaces/<slug>/prompt-labels`: `GET` list (member-readable; each label includes its nested `targets`), `POST` create, `PATCH`/`DELETE` `<label_id>`, and `POST`/`DELETE` `<label_id>/assignments` (attach/detach). Writes admin-gated via the existing `_is_workspace_admin` helper; attach targets validated against the registry/recipes.
- Contract tests: CRUD, attach/detach, idempotency, invalid-target validation, admin/member/outsider permissions, workspace scoping.

## Frontend (`apps/web`)

- `prompt-label` types, service, thin MobX store (mirrors `prompt-section`), hook, root-store wiring.
- Label chips on section/receipt cards, an admin attach/detach menu, a manage-labels modal (create/rename/recolor/delete with `ColorPicker`), and a top-of-page label bar whose chips filter the current tab.

## Validation

- Frontend: `check:types` clean for my code (one pre-existing, unrelated failure in `apps/web/tests/runners/runner-runs-page.test.tsx` re `IAgentRunPage`, identical to `main`); oxlint 0/0; oxfmt clean.
- Backend: `ruff` clean; `py_compile` OK. **pytest was not runnable in the authoring environment** (the API dependency tree isn't installed there); the contract tests run under `--nomigrations` in CI.

### DIFF
--- a/apps/api/pi_dash/prompting/migrations/0005_prompt_label.py
+++ b/apps/api/pi_dash/prompting/migrations/0005_prompt_label.py
@@ -1,0 +1,142 @@
+# Copyright (c) 2023-present Pi Dash Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
+"""Workspace-scoped prompt labels for sections and receipts (PDASHOSS01-71).
+
+Introduces ``PromptLabel`` (a workspace-shared, user-managed label) and
+``PromptLabelAssignment`` (the n:n attachment of a label to a code-defined
+prompt section — keyed by ``section_key`` — or receipt — keyed by prompt
+``kind``). Kept separate from the issue ``Label`` model, which is
+project/issue-coupled and hierarchical.
+"""
+
+import uuid
+
+import django.db.models.deletion
+from django.conf import settings
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
+        ("db", "0001_initial"),
+        ("prompting", "0004_prompt_section_override"),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="PromptLabel",
+            fields=[
+                (
+                    "id",
+                    models.UUIDField(
+                        default=uuid.uuid4,
+                        editable=False,
+                        primary_key=True,
+                        serialize=False,
+                    ),
+                ),
+                ("name", models.CharField(max_length=64)),
+                ("color", models.CharField(default="#6b7280", max_length=32)),
+                ("created_at", models.DateTimeField(auto_now_add=True)),
+                ("updated_at", models.DateTimeField(auto_now=True)),
+                (
+                    "created_by",
+                    models.ForeignKey(
+                        blank=True,
+                        null=True,
+                        on_delete=django.db.models.deletion.SET_NULL,
+                        related_name="prompt_labels_created",
+                        to=settings.AUTH_USER_MODEL,
+                    ),
+                ),
+                (
+                    "updated_by",
+                    models.ForeignKey(
+                        blank=True,
+                        null=True,
+                        on_delete=django.db.models.deletion.SET_NULL,
+                        related_name="prompt_labels_updated",
+                        to=settings.AUTH_USER_MODEL,
+                    ),
+                ),
+                (
+                    "workspace",
+                    models.ForeignKey(
+                        on_delete=django.db.models.deletion.CASCADE,
+                        related_name="prompt_labels",
+                        to="db.workspace",
+                    ),
+                ),
+            ],
+            options={
+                "db_table": "prompt_label",
+                "ordering": ("name",),
+            },
+        ),
+        migrations.CreateModel(
+            name="PromptLabelAssignment",
+            fields=[
+                (
+                    "id",
+                    models.UUIDField(
+                        default=uuid.uuid4,
+                        editable=False,
+                        primary_key=True,
+                        serialize=False,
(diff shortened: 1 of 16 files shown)


irichard00 (commented):
## Review — prompt-section/receipt labels (PDASHOSS01-71)

Reviewed the full diff (16 files, +1653/-10). This is a clean, well-architected full-stack feature that closely mirrors the existing `PromptSectionOverride` patterns. Verified locally:

- **Frontend `tsc --noEmit`**: the label code type-checks clean. The only error is the pre-existing, unrelated `tests/runners/runner-runs-page.test.tsx` `IAgentRunPage` cast — that file isn't in this PR and the error reproduces on `main`.
- **Backend**: `py_compile` OK on all changed Python. Migration `0005` mirrors the models (constraints, indexes, FKs, `db.workspace` FK dependency is correct — `Workspace` lives in `db/0001_initial`).
- **Contract tests**: read them against the shared fixtures — the role assumptions (`session_client`/`create_user` = admin role 20, `member_client` = role 15, `WORKSPACE_ADMIN_ROLE = 20`) all line up, and the workspace-scoping test is valid because `create_user` is a member of both workspaces. I could **not** run pytest here (the API dependency tree isn't installed in this environment); see note 3.

Nothing blocks merge from a code-correctness standpoint. The items below are one design decision for the maintainer to confirm and a few refinements.

---

### 1. (Decision needed) Label writes are admin-only, but the issue says "allow **user** add labels"

`PromptLabelListEndpoint.post`, `PromptLabelDetailEndpoint`, and `PromptLabelAssignmentEndpoint` all gate create/edit/delete/attach/detach behind `_is_workspace_admin`; the "＋ Label" attach menu and detach "×" in `PromptCardLabels` only render for `isAdmin`. Reads (list + click-to-filter) are open to any member.

This is the documented default (mirrors how section overrides are admin-gated) and was flagged in the issue thread with no reply. But it does diverge from the literal issue wording ("allow user add labels…"). This is the one thing worth an explicit maintainer call before merge:

- keep **admin-managed** (current), or
- let **any member** attach/detach (and possibly create) labels.

It's a small change either way — the role check is centralized — so it's cheap to flip.

### 2. Click-to-filter is per-tab; matches on the *other* tab are hidden

The issue asks that clicking a label "list out all associated sections **or** receipts." Right now `tab` and `activeLabelId` are independent page state, and each tab filters its own list. If you're on the **Sections** tab and click a label that's only attached to receipts, you get "No sections carry this label" and have to manually switch to the **Receipt** tab to see the matches. The bar's summary line ("… N sections, M receipts") hints that the other tab has matches, which softens it, but a user could reasonably think nothing matched. Consider auto-switching tabs when the active tab has zero matches but the other has some, or showing both in a single filtered view.

### 3. Contract tests are unverified in CI-equivalent form here — please confirm CI is green

I couldn't execute pytest locally (missing `corsheaders`/`psycopg`/etc.). The tests read correctly and cover CRUD, attach/detach, idempotency, invalid-target validation, admin/member/outsider permissions, and workspace scoping — but they've only been statically verified. Just confirm the `contract` job passes in CI before merge.

### 4. (Minor / hardening) `color` is stored without format validation

`POST`/`PATCH` accept any string up to 32 chars for `color` and it's written straight through. The UI only ever sends hex from `ColorPicker`, and it's rendered via a React style object (not raw CSS), so this isn't an injection risk — but a non-hex value would break the chip border, which is built by string concat: `borderColor: \`${label.color}66\`` in `PromptLabelChip`. A light server-side hex check (or documenting the contract) would make the API self-defending.

### 5. (Nit) Chip click is select-only on cards, but toggles on the bar

In `PromptLabelBar`, clicking the active chip clears the filter (`activeLabelId === label.id ? null : label.id`). On a card (`PromptCardLabels`), `onClick={() => onSelect(label.id)}` always sets and never clears. Minor inconsistency — clicking an already-active chip on a card is a no-op instead of toggling off. There is a "Clear filter" control, so this is cosmetic.

### 6. (Nit) Empty labels row on receipt cards for non-admins

`ReceiptCard` always renders the `flex … px-4 pb-3` wrapper around `PromptCardLabels`, which returns `null` when there are no labels and the viewer isn't an admin — leaving an empty padded strip. Trivial; only affects the no-labels non-admin case.

---

Overall: solid work, consistent with the codebase, safe to merge once (1) is confirmed and CI is green. Items 2/4 are worth a follow-up; 5/6 are nits.

PR #325, opened 2026-08-07
PDASHOSS01-78 Update "in progress" ticking cadence to 12 hours

irichard00:
## What

Lengthen the default periodic re-invocation ("ticking") cadence for the **In Progress** (started) phase from **3 hours to 12 hours**.

## How

- `Project.agent_default_interval_seconds` default raised `10800 → 43200` (the documented In Progress phase default; resolved through `IssueAgentTicker.effective_interval_seconds()`).
- New migration `0156_increase_in_progress_interval` performs an `AlterField` only — **no data migration**, mirroring the precedent in `0155_reduce_review_max_ticks`. Existing operator-customized project rows keep their stored value; only new projects pick up 12h.
- The **In Review** cadence (`agent_review_default_interval_seconds`) is intentionally unchanged.
- Prompt prose updates automatically: `_humanize_interval(43200)` → `"12 hours"`.

## Validation

- Verified `_humanize_interval(43200)` renders `"12 hours"`.
- Confirmed migration `0156` is the sole child of the current head `0155`, with the `AlterField` default matching the model default 1:1.
- Note: the full Django test suite could not be executed in the runner environment (no usable virtualenv present in the checkout). Existing ticker/context tests set the interval explicitly in fixtures, so a default change does not alter their assertions.

Resolves PDASHOSS01-78.

### DIFF
--- a/apps/api/pi_dash/db/migrations/0156_increase_in_progress_interval.py
+++ b/apps/api/pi_dash/db/migrations/0156_increase_in_progress_interval.py
@@ -1,0 +1,38 @@
+# Copyright (c) 2023-present Pi Dash Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
+"""Increase the default In Progress ticking cadence from 3 h to 12 h.
+
+Per PDASHOSS01-78 the In Progress phase re-invokes the agent too often at
+the 3 h interval; the cadence is lengthened to 12 h (43200 s) so periodic
+runs are spaced further apart. Only the In Progress phase changes — the
+In Review cadence (``agent_review_default_interval_seconds``) is untouched.
+
+Only the project-level default changes; per-issue
+``IssueAgentTicker.interval_seconds`` overrides are untouched. Following
+the same rationale as ``0155_reduce_review_max_ticks`` a data migration is
+intentionally *not* run: projects created before this change already
+materialized ``10800`` into their row, and silently rewriting an
+operator's stored value is more surprising than leaving it. New projects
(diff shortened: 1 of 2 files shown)


irichard00 (commented):
**Review: approved — no changes requested.**

Verified against the codebase:

- **Right source of truth.** The In Progress cadence resolves through `IssueAgentTicker.effective_interval_seconds()` (`issue_agent_ticker.py:150`). The non-review branch reads `project.agent_default_interval_seconds`; the review branch reads the separate `agent_review_default_interval_seconds`. Changing only the former lengthens In Progress alone, leaving In Review at 3h — matches the issue intent.
- **Migration graph clean.** `0156` depends on `0155_reduce_review_max_ticks` (the current head) and is its sole child. The `AlterField(default=43200)` matches the model default 1:1.
- **Precedent followed.** No data migration, mirroring `0155`: operator-customized rows keep their stored value; only new projects get 12h. Per-issue `interval_seconds` overrides are untouched.
- **Prose renders correctly.** `_humanize_interval(43200)` → `"12 hours"` (43200 % 3600 == 0, 43200 // 3600 == 12).
- **No test regressions.** The ticker/context tests set `agent_default_interval_seconds` explicitly in the `project_with_overrides` fixture (`test_issue_agent_ticker.py:32`), so their `== 10800` assertions depend on the fixture, not the model default.

**Caveat (unchanged from author):** the full Django suite can't run in this checkout — app deps (`corsheaders`, DRF, etc.) aren't installed in the available interpreter, so `pytest` / `makemigrations --check` couldn't be executed. Verification above is by static analysis. Recommend confirming `makemigrations --check` is clean in CI before merge.

PR #327, opened 2026-08-14
feat(assistant): user-configurable MCP tool servers

irichard00:
Adds a generic MCP client to the assistant. Users register streamable-HTTP tool servers in settings, and those tools become available to the agent during every run, alongside the built-in ones.

This is a standalone feature for CE users — nothing in it is specific to any deployment. It also introduces the `resolve_toolsets_for_user` seam, mirroring the existing `resolve_model_for_user`, so a downstream build can inject additional toolsets without patching the runtime.

## What's here

- **`AssistantMCPServer`** — per-user, unique name, optional `Authorization` header encrypted at rest with the same backend as BYOK provider keys.
- **`runtime/mcp.py`** — turns rows into pydantic-ai toolsets. Uses `MCPToolset` rather than the deprecated `MCPServerStreamableHTTP` (removed in pydantic-ai v2).
- **The seam** in `ee/assistant/model_provider.py`, and `toolsets=` wired into `_run_turn`.
- **CRUD API** under `/api/users/me/ai-assistant/mcp-servers/`, reusing the existing SSRF policy for user-supplied URLs.
- **Settings UI** section under AI Assistant.

## Design note: tool servers are additive

A tool server must never be able to take down the assistant. A server that is blocked, unreachable, or has an undecryptable credential is **skipped for that run** — the turn proceeds with whatever else built, and a `tool_servers_skipped` event tells the user which one was dropped and why. Silently losing a capability is nearly as bad as failing, so the event matters as much as the fail-open.

`tool_prefix` is derived from the server name and always non-empty (falling back to the row id), since an empty prefix would let a server's tools collide with the built-in ones.

## Test plan

- [x] `pi_dash/tests/contract/assistant/test_mcp_servers.py` — model, CRUD, and toolset construction, including: credential never echoed back, preserved across an unrelated PATCH, clearable with an explicit empty string; ownership scoping; per-user name uniqueness; blocked URLs rejected on create and skipped (not fatal) at build time; one broken row not affecting the others
- [ ] CI run (the ambient dev environment here has no `celery`, so the suite could not be executed locally)


irichard00:
## Self-review finding (fixed in 832073bd)

One finding, surfaced while reviewing the downstream cloud PR.

**Credit exhaustion was indistinguishable from a crash.** `_classify_error` had no branch for 402, so a provider rejecting a request for lack of balance produced *"The assistant hit an unexpected error."* — and logged it via `logger.exception` as though it were a bug.

This isn't cloud-specific: OpenRouter and any other routed or metered provider answers 402 for an empty balance, so a CE user on BYOK hits it too. Added a `provider_out_of_credit` classification alongside the existing auth/unreachable/model branches.

Also worth a reviewer's eye: **`MCPServerStreamableHTTP` is deprecated** and removed in pydantic-ai v2, so this uses `MCPToolset` with `.prefixed()` — prefixing moved from a constructor argument to a wrapper, and the connect timeout is `init_timeout`. Worth knowing if anyone writes MCP code against the older docs.

irichard00:
## Second review pass — one serious finding (fixed in 7f7368ed)

Went back over the resilience claim rather than trusting it, and it was **half-implemented**.

### A dead tool server failed the entire turn

The docstring says *"a tool server that is unreachable must never take the assistant down with it"* — but `build_toolsets` only skipped failures during **construction**, and construction does no I/O. The connection opens when the agent *enters* the toolset, at the start of the run. Verified directly against a refused port:

```
RESULT: run RAISED -> RuntimeError  Client failed to connect: All connection attempts failed
```

So one broken tool server cost the user their assistant entirely — the exact failure the design set out to prevent. My own comment ("Building a toolset never performs I/O — an unreachable server fails later, during the run") described the hole accurately without my noticing it *was* one.

`ResilientToolset` now absorbs connect and tool-listing failures, reports no tools, and records the reason; `_report_runtime_tool_failures` emits `tool_servers_skipped` afterwards so the user still learns which capability was missing. Re-verified: the run now completes with the failure recorded.

### Colliding tool prefixes silently shadowed each other

Names are unique per user, but slugification is lossy — "My Tools", "my-tools" and "my_tools" all reduce to `my_tools`. Two servers sharing a prefix expose duplicate tool names to the model, which has no way to detect it. Prefixes are now disambiguated per run, ordered by `created_at` so they stay stable.

Both are covered by new tests, including one that runs a real agent turn against a refused port.

irichard00:
## Third review pass (2f2f0bcf)

Adds one seam, prompted by tracing the downstream cloud build against its design goals.

`_model_label` read the BYOK config directly to decide what to record as a turn's `model_used`. Any build that resolves models from somewhere other than that config — which is what the `resolve_model_for_user` seam exists to allow — therefore recorded an **empty** model on every turn.

That is a gap in the seam set rather than a cloud problem: `resolve_model_for_user` lets a build choose the model, but nothing let it say what it chose. `model_label_for_user` closes it. **CE behaviour is unchanged** — the default implementation is exactly the code that was inlined before, moved behind the seam.

PR #326, opened 2026-08-07
OPENHUB-45 Add folded comments concept to PiDash

irichard00:
## Summary

- add opt-in `fold` labels to issue comments and expose them through the API/types
- add `--fold` to `pidash comment add` and `fold` to the assistant comment tool
- exclude folded comments from agent-run prompt context and document the behavior for agents
- render folded comments collapsed by default with click-to-expand in Web and Space

## Validation

- `cargo test --manifest-path runner/Cargo.toml cli::comment` (6 passed)
- feature-specific Django prompt/assistant/serializer tests (5 passed)
- `manage.py sqlmigrate db 0156`
- `pnpm turbo run check:types --filter=space` and dependency builds
- changed-file OxLint and formatting checks

## Existing baseline issues observed

- Web type checking reaches an unrelated existing fixture error in `apps/web/tests/runners/runner-runs-page.test.tsx:94` (`IAgentRunPage` fixture lacks `page`/`per_page`); no changed file reports a type error.
- An expanded prompting test run passes 41/42; the remaining existing soft-delete test tries to connect to RabbitMQ on an unexposed host port.
- `makemigrations --check` reports unrelated pre-existing model drift after migration `0156`; it proposes no further change for `IssueComment.labels`.

PR #330, opened 2026-08-18
Make folded comments visible as a count in the agent prompt

irichard00:
## Summary

Follow-up to #326. Folded comments were removed from the run prompt *silently* — the agent could not tell anything had been hidden, so it had no basis for deciding whether to go read it.

Worse, an issue whose comments were all folded rendered as `(no comments on this issue yet)`, which is false and could lead an agent to conclude a thread had never been discussed.

- `_comments_section` appends a note with the folded count and the command to read them
- the all-folded case now says so explicitly instead of claiming the thread is empty
- the parent-issue line gains the same subset count: `The parent has 7 comment(s), 1 of them folded; ...`
- `validation.py`'s sample context gains `folded_comments_count` so the strict renderer keeps passing for every kind

Folded comments remain excluded from the prompt body — only their existence is surfaced. `pidash comment list` is unchanged.

## Validation

- `pytest pi_dash/tests/unit/prompting` — 95 passed, 1 pre-existing failure (`test_context_code_reviews_excludes_soft_deleted` needs RabbitMQ, unexposed locally; same baseline noted in #326)
- Three new tests: folded note present, all-folded thread does not read as empty, unfolded thread carries no note
- Rendered the parent line at folded counts 0/1/3 to confirm the clause only appears when non-zero

## Note

`fragments/01_intro.md` is updated alongside `sections/intro.md` per the sync convention established in #264 / #272. Only `sections/` reaches a real run.

### DIFF
--- a/apps/api/pi_dash/prompting/context.py
+++ b/apps/api/pi_dash/prompting/context.py
@@ -52,6 +52,19 @@ def _issue_comment_count(issue: Issue) -> int:
     return IssueComment.objects.filter(issue=issue).count()
 
 
+def _folded_comment_count(issue: Issue) -> int:
+    """How many of ``issue``'s comments carry the ``fold`` label.
+
+    Folded comments are omitted from the rendered thread (see
+    ``_comments_section``), so the count is surfaced alongside it to keep the
+    omission visible — an agent that cannot see *that* something was hidden
+    has no way to decide whether it needs to go read it.
+    """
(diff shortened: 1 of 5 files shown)


irichard00:
Closing — the existing prompt already gives the agent what it needs here: visible comments are numbered in the rendered thread, and the parent line carries its own count. Not worth the extra prompt surface. Branch `pi-dash/openhub-45-folded-counts` is left in place if we want to revisit.

PR #333, opened 2026-08-21
PDASHOSS01-82 MCP connector: include the issue web URL in issue payloads

irichard00:
## What & why

No MCP tool response (or `pidash` CLI issue payload) carried a clickable link to the issue in the web UI — only UUIDs and sequence ids. An agent that had just filed work for a human couldn't hand them a link. This adds an absolute, human-clickable `url` to every issue-returning payload.

## Approach

There is **no standalone MCP server in this repo** — the hosted claude.ai "Pi Dash" connector and the `pidash` CLI both wrap this repo's REST `/api/v1/` surface, and the CLI prints the REST payload verbatim. So adding `url` to the REST serializers propagates to **both** the MCP tools and the CLI.

- **`utils/host.py`** — `web_base_url()` + `issue_web_url(slug, project_identifier, sequence_id)`. Built from deployment config (`WEB_URL`, falling back to `APP_BASE_URL`), **never** the inbound request host (which may be a proxy / non-public hostname). Returns `None` when unconfigured.
- **`IssueSerializer` / `IssueCommentSerializer`** — add a `url` field using the canonical browse route `…/<slug>/browse/<PROJ>-<seq>` (the UI resolves this directly; the UUID route redirects to it). **Omitted** from the payload when no base URL is configured — a missing `url` is easy for a client to handle, a broken one is not. Covers `get` / `create` / `update` / `list` / `move` / `comment` tools + CLI `issue get`.
- **Advanced search endpoint** — each result row's `url` is now the absolute web URL (previously a relative API path), omitted when unconfigured.
- **Docs** — new `docs/mcp-connector.md` documents the field; CLI reference updated.

## Testing

- New unit tests: `tests/unit/utils/test_web_url.py`, `tests/unit/serializers/test_issue_url.py`.
- Extended `tests/contract/api/test_issue_search.py` with web-URL present/omitted cases.
- Full `tests/contract/api/` (133) + `tests/unit/serializers/` + `tests/contract/assistant/` (78) pass; `ruff check` clean.

## Acceptance criteria

- [x] Every issue-returning MCP tool includes an absolute `url`
- [x] Value built from configured deployment config, omitted when unset
- [x] A link from a `create_issue` response opens the issue in the web UI (canonical browse route)
- [x] `docs/mcp-connector.md` documents the field
- [x] CLI `pidash issue get` includes the `url` field

### DIFF
--- a/apps/api/pi_dash/api/serializers/issue.py
+++ b/apps/api/pi_dash/api/serializers/issue.py
@@ -36,6 +36,7 @@ from pi_dash.utils.content_validator import (
     validate_html_content,
     validate_binary_data,
 )
+from pi_dash.utils.host import issue_web_url
 
 from .base import BaseSerializer
 from .cycle import CycleLiteSerializer, CycleSerializer
@@ -85,6 +86,12 @@ class IssueSerializer(BaseSerializer):
     type_id = serializers.PrimaryKeyRelatedField(
         source="type", queryset=IssueType.objects.all(), required=False, allow_null=True
     )
+
+    # Absolute, human-clickable web URL for the issue. Built from deployment
+    # config (see ``issue_web_url``); omitted from the payload when no web base
+    # URL is configured (a missing url is easy for a client to handle, a broken
+    # one is not).
+    url = serializers.SerializerMethodField()
 
     class Meta:
         model = Issue
@@ -337,8 +344,19 @@ class IssueSerializer(BaseSerializer):
         instance.updated_at = timezone.now()
         return super().update(instance, validated_data)
 
+    def get_url(self, obj):
+        return issue_web_url(
+            obj.workspace.slug if obj.workspace_id else None,
+            obj.project.identifier if obj.project_id else None,
+            obj.sequence_id,
+        )
+
     def to_representation(self, instance):
         data = super().to_representation(instance)
+        # Omit ``url`` entirely when no web base URL is configured, rather than
+        # emitting ``null`` — keeps the contract "present and absolute, or absent".
+        if data.get("url") is None:
+            data.pop("url", None)
         if "assignees" in self.fields:
             if "assignees" in self.expand:
                 from .user import UserLiteSerializer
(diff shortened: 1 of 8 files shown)


irichard00 (commented):
## Review — approve (no blocking findings)

Reviewed the full diff against the acceptance criteria. The implementation is clean, well-scoped, and correct.

**Validated against acceptance criteria**

- **URL shape is canonical and actually resolves.** `issue_web_url()` builds `<base>/<slug>/browse/<PROJ>-<seq>`. I checked this against the frontend, not just the tests: `apps/web/app/routes/core.ts:86` defines `:workspaceSlug/browse/:workItem`, and the app's own UUID-route page (`.../issues/(detail)/[issueId]/page.tsx:32`) redirects to `/${workspaceSlug}/browse/${project_identifier}-${sequence_id}` — the *identical* string this helper emits. So a `url` taken straight from a `create_issue` response lands exactly where the app itself would send the user.
- **Base URL comes from deployment config, never the request host** — `settings.WEB_URL or settings.APP_BASE_URL`, mirroring the existing `base_host()` convention. ✅
- **Omitted (not null, not relative) when unconfigured** — handled in three places consistently: `IssueSerializer.to_representation` pops `url` when `None`, `IssueCommentSerializer.to_representation` likewise, and the advanced-search view only sets `result["url"]` when the helper returns a value. `web_base_url()` also correctly treats blank strings as unset. ✅
- **Comment payloads point at the parent issue** (no per-comment anchor route) — matches the UI, documented. ✅
- **CLI coverage** — `pidash issue get` prints the REST payload verbatim, so `url` flows through with no Rust change; documented in `docs/wiki/17-cli-reference.md`. ✅
- **Docs** — `docs/mcp-connector.md` added, documents the field, shape, config source, and omission behavior. ✅

**Test coverage** is solid: helper (present / trailing-slash / fallback / unset / blank / missing-parts), both serializers (present + omitted), and the search endpoint (absolute present + omitted). Independently confirmed `IssueSerializer.Meta` uses `exclude`, so the declared `SerializerMethodField` is included in output.

**Non-blocking nits** (no change requested):

- The advanced-search `url` field changes meaning from a relative *API* path to an absolute *web* URL. That's the right call per the acceptance criteria (uniform clickable `url`), and I found no in-repo consumer relying on the old API-path semantics — noting it only as a behavior change for any external client.
- On single-object `create`/`update` responses, `get_url` may trigger one extra query to load `workspace.slug` / `project.identifier` if the instance wasn't `select_related`. Negligible (one row, one write path); list/get/search paths already prefetch.

Scoping of `pidash_list_projects` / `pidash_get_workpad` URLs as follow-ups is reasonable — the ticket lists them as "worth considering," and keeping this PR to issue payloads is the right boundary.

PR #335, opened 2026-08-21
PDASHOSS01-87 Adding AI reasoning complexity score to the pi dash issue

irichard00:
## Summary

Adds an integer `complexity_score` metadata field to the Issue — an AI-reasoning-complexity rating used to route an issue to an LLM of matching capability.

- `0` means **"no score"** (unrated) and is the **default**.
- A rated issue takes an integer **1..10**.
- Values outside `0..10` are rejected (rule check).

Deriving/rating the score is intentionally **out of scope** for this change (per the issue) — this only adds the field, its bounds, and API exposure.

## Changes

- **Model** (`db/models/issue.py`): `complexity_score = IntegerField(default=0, validators=[MinValueValidator(0), MaxValueValidator(10)])`.
- **Migration** `0159_issue_complexity_score` — additive; existing issues take the `0` default.
- **Serializers** (`app/serializers/issue.py`): exposed on `IssueSerializer` and `IssueFlatSerializer`; auto-included on the write path (`IssueCreateSerializer` uses `exclude`). Added `validate_complexity_score` for a clear, field-scoped rule-check error message.
- **Porter serializer** (`utils/porters/serializers/issue.py`): added to the export field list.
- **Tests** (`tests/unit/serializers/test_issue_complexity_score.py`): default is 0; accepts 0/1/5/10; rejects -1/11/100; save round-trips.

## Validation

- Targeted test: 9/9 pass (incl. DB round-trip proving the column).
- Regression: `tests/unit/serializers/` + `test_issue_assigned_pod.py` = 58/58 pass.
- `makemigrations db --check`: `0159` captures the field; no missing migration for `complexity_score`.

## Follow-ups (not in this PR)

`IssueVersion` snapshot, activity-history tracking, frontend TS types / UI control, and CSV exporter column are left for separate triage.

### DIFF
--- a/apps/api/pi_dash/app/serializers/issue.py
+++ b/apps/api/pi_dash/app/serializers/issue.py
@@ -98,6 +98,7 @@ class IssueFlatSerializer(BaseSerializer):
             "description_json",
             "description_html",
             "priority",
+            "complexity_score",
             "start_date",
             "target_date",
             "sequence_id",
@@ -173,6 +174,19 @@ class IssueCreateSerializer(BaseSerializer):
         label_ids = self.initial_data.get("label_ids")
         data["label_ids"] = label_ids if label_ids else []
         return data
+
+    def validate_complexity_score(self, value):
+        # AI-reasoning complexity rule check: only 0..10 is allowed. 0 means
+        # "no score" (unrated); a rated issue is 1..10. The model already carries
+        # Min/Max validators, but re-checking here gives a precise, field-scoped
+        # error instead of DRF's generic min/max wording.
+        if value is None:
+            return value
+        if not 0 <= value <= 10:
+            raise serializers.ValidationError(
+                "complexity_score must be an integer from 0 to 10 (0 means unrated)."
+            )
+        return value
 
     def validate(self, attrs):
         allow_triage = self.context.get("allow_triage_state", False)
@@ -991,6 +1005,7 @@ class IssueSerializer(DynamicBaseSerializer):
             "completed_at",
             "estimate_point",
             "priority",
+            "complexity_score",
             "start_date",
             "target_date",
             "sequence_id",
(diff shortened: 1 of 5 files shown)


irichard00:
## Code review — `complexity_score` (PDASHOSS01-87)

**Verdict: approve.** The change is minimal, additive, and satisfies every acceptance criterion. No blocking issues found.

### What I verified
- **Model** — `Issue.complexity_score = IntegerField(default=0, validators=[Min(0), Max(10)])`. Default `0` = unrated; rated is `1..10`. ✔
- **Migration** `0159_issue_complexity_score` — additive `AddField`, depends on `0158_test_cadence_fields` (the true head on `main`; no numbering collision). No data migration needed — existing rows take the `0` default. ✔
- **Write path** — `issue/base.py:208` routes `create`/`update`/`partial_update` to `IssueCreateSerializer`, which uses `exclude=` so the field is auto-included, plus an explicit `validate_complexity_score` for a precise rule-check message. Both create and PATCH are covered. ✔
- **Read / export** — exposed on `IssueSerializer`, `IssueFlatSerializer`, and `IssueExportSerializer`. GET returns the value. ✔
- **Tests** — `test_issue_complexity_score.py`: default-is-0, accepts 0/1/5/10, rejects -1/11/100, DB round-trip. **Ran the suite independently in the Docker harness: 9/9 pass; serializer dir 55/55, no regressions.** ✔

### Notes (non-blocking, already flagged as out of scope by the author)
The issue scoped this to "add the field"; the rating logic is explicitly out of scope. These are reasonable follow-ups, not gaps in this PR:
- `IssueVersion` snapshot does not carry `complexity_score`.
- Activity-history tracking (`bgtasks/issue_activities_task.py` `track_priority` pattern) does not track changes to this field.
- No frontend TS type / UI control.
- CSV **import** path not touched (export column added).

The `if value is None: return value` guard in `validate_complexity_score` is dead-but-harmless — the non-nullable model field means DRF rejects `null` before the validator runs. Fine to leave.

Ready to merge from my side.

PR #338, opened 2026-08-25
PDASHOSS01-75 Prevent duplicate runner working directories

irichard00:
## What

Runners created from the cloud/UI go through the daemon's `create_runner_inner`, which hardcoded `RunnerWorkdirPlan::Legacy`. A legacy runner runs directly in `workspace.working_dir` and is **not** exempt from `Config::validate()`'s `DuplicateWorkingDir` check, so creating a second UI runner on a repo already used by another runner failed with:

> Each runner must have its own working directory; concurrent git operations on the same tree corrupt state silently.

Per the issue decision, legacy single-dir runner creation is deprecated — our own code should no longer create legacy runners (the legacy code stays only for backward compatibility with un-upgraded CLI installs).

## Changes

- **`machine_control.rs`** — `create_runner_inner` now uses `RunnerWorkdirPlan::AutoPoolIfGit` (the same plan `pidash runner add` defaults to). On a git repo it creates/reuses a `[[workdir]]` pool and migrates any existing exact-path legacy runner into it, so two runners can share one repo via leased worktrees and both are exempt from the duplicate-dir check.
- **`supervisor.rs`** — worktree pools were built once at daemon startup, so a runner hot-added afterward would reference a pool that isn't live and fail every run with *"work dir has no worktree pool"* until the next restart. The shared pool map is now `Arc<RwLock<HashMap<..>>>`, and `RunnerSpawnCtx::add_runner` builds and registers the runner's pool on hot-add when it isn't already live — so UI-created runners work immediately, no restart required.
- **`ipc/server.rs`** — pool map field/reader updated for the new `RwLock` type; `pidash status` still snapshots live pools.

`RunnerWorkdirPlan::Legacy` and its code paths are retained for backward compatibility.

## Testing

- `cargo build -p pidash` ✅
- `cargo test -p pidash --lib` ✅ (455 passed; the one failure `workspace::git::tests::checked_out_branches_...` is a pre-existing macOS `$TMPDIR` `/private/var` vs `/var` symlink mismatch — reproduces identically on `main` with this change stashed, unrelated)
- All integration test targets compile (`cargo test -p pidash --no-run`)

### DIFF
--- a/runner/src/daemon/machine_control.rs
+++ b/runner/src/daemon/machine_control.rs
@@ -220,18 +220,23 @@ impl MachineControl {
         .await
         .map_err(|e| anyhow::anyhow!("cloud registration failed: {e}"))?;
 
-        // 2. Persist the `[[runner]]` block. `Legacy` workdir plan —
-        //    worktree pools are built once at daemon startup, so a
-        //    hot-added runner can't join one until the next restart;
-        //    running the agent directly in working_dir keeps config and
-        //    runtime consistent. Operators can migrate via the CLI.
+        // 2. Persist the `[[runner]]` block. `AutoPoolIfGit` — the same
+        //    plan `pidash runner add` defaults to: if `working_dir` is a
+        //    git repo, create or reuse a `[[workdir]]` pool and bind this
+        //    runner to it (migrating any existing legacy runner on the
+        //    exact same path). Legacy single-dir runners are deprecated and
+        //    no longer created by our own code; the pool exemption is what
+        //    lets two runners share one repo without a `DuplicateWorkingDir`
+        //    collision. Worktree pools are otherwise built once at daemon
+        //    startup, so `add_runner` below builds and registers this
+        //    runner's pool on hot-add — no restart required.
         let options = ApplyEnrollOptions {
             working_dir: (!cmd.working_dir.is_empty()).then(|| PathBuf::from(&cmd.working_dir)),
             agent_kind,
             model: (!cmd.model.is_empty()).then_some(cmd.model.as_str()),
             reasoning_effort: (!cmd.reasoning_effort.is_empty())
                 .then_some(cmd.reasoning_effort.as_str()),
-            workdir_plan: RunnerWorkdirPlan::Legacy,
+            workdir_plan: RunnerWorkdirPlan::AutoPoolIfGit,
         };
         let cloud_url = self.spawn_ctx.cloud_url();
         let applied = match apply_enroll_response(&self.spawn_ctx.paths, &resp, &cloud_url, options)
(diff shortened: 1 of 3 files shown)


irichard00:
## Review — PDASHOSS01-75 (auto review pass)

Reviewed commit `ee69e0b3` (3 files, +80/−27). The change does what the issue asks: the daemon/UI creation path (`create_runner_inner`) now uses `AutoPoolIfGit` instead of hardcoded `Legacy`, and `add_runner` builds+registers the pool on hot-add so the new runner can run without a daemon restart. Verified statically that **all five** `.pools` access sites and both construction sites were migrated to `Arc<RwLock<HashMap<..>>>` consistently, the two lock scopes are correct (the `read()` guard in `add_runner` is bound to a `bool` and dropped before `ensure_pool` takes the write lock — no self-deadlock; the guard in `spawn_instance_tasks` is scoped to a block and dropped before the spawns). The Legacy variant is retained for backward compat. Nice, tightly-scoped fix.

**Approving with three non-blocking notes** — none break the happy path.

### 1. `ensure_pool` comment claims a re-check it doesn't do (minor)
`supervisor.rs` ~L409:
```rust
// Re-check under the write lock: a concurrent hot-add may have raced
// us. Last writer wins; a redundant pool's owner task idles harmlessly.
self.pools.write().await.insert(name.to_string(), handle);
```
The insert is unconditional — there is no re-check. So on a concurrent hot-add of the *same* workdir this overwrites whatever `PoolHandle` is already in the map (dropping it) rather than dropping the loser's freshly-spawned handle. The `already_live` guard in `add_runner` makes this reachable only on a genuine race, so impact is low, but consider matching code to comment: `if !self.pools.write().await.contains_key(name) { insert }` (or `entry(..).or_insert`) and let the redundant handle drop. Also worth confirming that a dropped `PoolHandle` actually tears down its owner task (the comment asserts "idles harmlessly").

### 2. Migration degrade window (design note — already documented in the test handoff)
When a *pre-existing* legacy runner shares the path, `AutoPoolIfGit` migrates it on disk but it keeps in-memory legacy behavior (runs directly in `working_dir`) until the next restart. During that window it runs concurrently with the new pooled runner on the same repo — i.e. the exact "concurrent git ops on the same tree" the issue warns about becomes briefly possible. Acknowledged as a known limitation; flagging for reviewer visibility since it's the one scenario where the fix doesn't fully close the original hazard.

### 3. No unit test for the new hot-add path (test-coverage)
`two_runners_share_one_workdir_concurrently` covers the pool concept but nothing exercises `ensure_pool`/`add_runner` registering a pool at runtime — the core new behavior and acceptance criterion "runs immediately after creation, no restart." A focused test (build `Paths`, write a config with one git `[[workdir]]`, call `ensure_pool`, assert the pool is registered) would lock this in cheaply without the full daemon scaffolding.

PR #339, opened 2026-08-25
PDASHOSS01-81 Pod edit and delete support in pidash

irichard00:
## What

Wires the missing **pod edit & delete UI** into the Runners page. The backend `PATCH`/`DELETE` endpoints (`PodDetailEndpoint`) and the `PodService.update`/`.remove` client already existed; the web UI only exposed create + list. Issue PDASHOSS01-81.

## Changes

- **`EditPodModal`** (`apps/web/core/components/runners/edit-pod-modal.tsx`) — rename, edit description, and promote a pod to the project default. Mirrors `CreatePodModal`. Sends only changed fields; strips/re-adds the `{PROJECT}_` name prefix like create does; disables un-defaulting the current default pod (the server only transfers the flag by promoting another pod).
- **Per-pod overflow menu** on each pod tile (`runners/page.tsx`) — an Edit / Delete `CustomMenu`. The filter tile was restructured from a bare `<button>` into a container so the menu button isn't nested inside it.
- **Delete confirm** via `AlertModalCore`, mapping the server's 409 guards (`pod_has_runners`, `pod_has_active_runs`, `default_pod_undeletable`) to plain-language toasts, and clearing the pod filter when the selected pod is deleted.
- **Tests** (`tests/runners/edit-pod-modal.test.tsx`) — prefill, no-op close, changed-field payloads, default promotion, disabled toggle, and error handling.

## Testing

- `vitest run tests/runners/edit-pod-modal.test.tsx tests/runners/create-pod-modal.test.tsx` → 18 passed.
- `tsc --noEmit` → no new errors in touched files.
- `oxlint` + `oxfmt --check` → clean on touched files.

Permission gating stays server-side (workspace admin or pod creator): the menu is shown to all members and the server's 403/409 responses surface as toasts, matching the existing runner Delete/Revoke pattern on this page.

### DIFF
--- a/apps/web/app/(all)/[workspaceSlug]/runners/page.tsx
+++ b/apps/web/app/(all)/[workspaceSlug]/runners/page.tsx
@@ -6,17 +6,19 @@
 
 import { useState } from "react";
 import { observer } from "mobx-react";
-import { HelpCircle, Plus } from "lucide-react";
+import { HelpCircle, MoreHorizontal, Pencil, Plus, Trash2 } from "lucide-react";
 import { useNavigate } from "react-router";
 import useSWR from "swr";
 import { useTranslation } from "@pi-dash/i18n";
+import { IconButton } from "@pi-dash/propel/icon-button";
 import { TOAST_TYPE, setToast } from "@pi-dash/propel/toast";
 import { PodService, RunnerService } from "@pi-dash/services";
 import type { IPod, IRunner } from "@pi-dash/types";
-import { AlertModalCore, Badge, Button, Tooltip } from "@pi-dash/ui";
+import { AlertModalCore, Badge, Button, CustomMenu, Tooltip } from "@pi-dash/ui";
 import { PageHead } from "@/components/core/page-title";
 import { AddRunnerModal } from "@/components/runners/add-runner-modal";
 import { CreatePodModal } from "@/components/runners/create-pod-modal";
+import { EditPodModal } from "@/components/runners/edit-pod-modal";
 import { RUNNER_STATUS_I18N_LABELS, STATUS_BADGE_VARIANT } from "@/components/runners/runner-status";
 import { RunnersTabs } from "@/components/runners/runners-tabs";
 import { useSelectedPodFilter } from "@/hooks/use-selected-pod-filter";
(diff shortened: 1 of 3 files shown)


irichard00:
## Review: Pod edit & delete UI — approved ✅

Reviewed the diff (3 files, +493/-17) against the acceptance criteria and the existing server contract. It's clean and well-scoped; no blocking issues.

**Verified**
- **Error shapes match the backend.** `PodService.update`/`.remove` re-throw `e.response.data`, and `PodDetailEndpoint` returns `{error, code}` on the 409 guards and `{error: "forbidden"}` on 403. The delete handler's `(err?.code && byCode[err.code]) || err?.error || fallback` maps all three guard codes (`pod_has_runners`, `pod_has_active_runs`, `default_pod_undeletable`) and correctly surfaces the raw `forbidden` message for 403 — no silent failure. A total network failure (no `response`) throws `undefined`, which safely lands on the generic fallback toast in both handlers.
- **Change-only PATCH payload.** `EditPodModal` diffs name-suffix / description / `is_default` and omits unchanged fields, so a no-op save closes without a network call, and the default toggle never re-issues a promote/demote no-op.
- **Default-pod invariant respected.** The toggle is `disabled` when the pod is already default, so the UI can never send `is_default: false` — matching the server, which only transfers the flag by promoting another pod.
- **Prefix handling.** `suffixOf()` strips `{project_identifier}_` for display and sends the bare suffix; the server re-prefixes and re-validates via `validate_user_pod_name`.
- **Valid DOM / a11y.** The filter tile was restructured from a `<button>` into a `<div>` with the filter button and the `CustomMenu` trigger as siblings (no nested-button), the trigger carries an `aria-label`, and `pr-9` reserves room so the menu doesn't overlap the label.
- **Filter cleared on delete** when the deleted pod is the active filter.
- **Tests:** ran `vitest run tests/runners/edit-pod-modal.test.tsx` locally → **10/10 pass** (prefill, whitespace-name block, no-op close, changed-field-only payload, default promotion, disabled toggle, backend-error + generic-fallback toasts).

**Non-blocking nits** (optional, fine to merge as-is)
- `AlertModalCore` for delete uses `primaryButtonText={{ default: t("Delete"), loading: t("Delete") }}` — the loading label is identical to the idle one, so there's no visible progress cue while the delete is in flight (the button is still disabled via `isSubmitting`). Consider a `t("Deleting…")` loading label to match the edit modal's `Saving…`.
- Server-side guard order (unchanged by this PR): a *default* pod that also has runners returns `pod_has_runners` before `default_pod_undeletable`, so test step 7 only shows the `default_pod_undeletable` toast when that pod has zero active runners. Worth knowing when validating manually; not a UI defect.

Approving. Recommend a manual pass against a live API for the delete-guard toasts (not exercisable in unit tests) before closing.

PR #343, opened 2026-08-29
PDASHOSS01-96 Project-scoped AI Workers panel; move workspace AI Workers into More

irichard00:
## Summary

Adds a **project-scoped AI Workers (runners) panel** at `/:workspaceSlug/projects/:projectId/runners` (+ `runs` / `approvals` / `chat/:runnerId` / `detail/:runnerId` children), reusing the exact same middle-panel page modules as the workspace-wide panel, filtered to one project. Demotes the workspace-wide "AI Workers" nav entry into the sidebar **More** section as the aggregate. Implements PDASHOSS01-96 (sub-issue of PDASHOSS01-95).

### Reuse, not fork
- The five panel pages (`overview`, `runs`, `approvals`, `chat`, `detail`) are referenced by **both** route trees via explicit route ids in `core.ts`. Each reads an optional `projectId` from the route and threads it into every list's service call + SWR key (so paging never leaks across scopes) and rewrites its link base. Workspace routes pass nothing → today's behaviour unchanged.
- The runner contacts sidebar is extracted into one shared `RunnersSideNav` used by both the workspace layout and the new thin project layout (`AppHeader` + `ContentWrapper` + shared side-nav + `Outlet`), mirroring the Schedulers dual-scope precedent.

### Backend filtering
- `runs`, `approvals`, and `chat-session` list endpoints gain an optional `?project=<uuid>` filter, narrowing to every row whose pod belongs to the project (a project owns several pods → filter by pods-of-project). The runners list already had `?pod=`; it gains `?project=` alongside.
- On the runs list the project filter is a **separate chained `.filter()`** so it AND-combines with the private-runner visibility gate rather than widening it — runs on another user's private runner stay hidden regardless of scope.

### Nav
- New project "AI Workers" sidebar item (sibling of Work Items / Schedulers, `Bot` icon, project ADMIN/MEMBER).
- Workspace `runners` entry moved from the top-level pinned list into **More**.

### Empty state
- `RunnersSideNav` shows a "No runners connected to this project yet." hint — the common case for Cloud-Agent-only projects.

## Tests
- New `test_runners_project_filter.py`: project-filter coverage for runs (incl. **private-runner rule preserved**), approvals, runners, and chat sessions. 5 passed.
- Existing runner view suites re-run green (77 passed).
- Web `check:types` + `check:lint` clean.

## Access decision
Project panel gates on project **ADMIN/MEMBER** (matching the workspace panel's ADMIN/MEMBER rule, applied at project scope). Flagged in the issue as an open question; documented here for review.


irichard00 (commented):
## Review — PDASHOSS01-96 Project-scoped AI Workers panel

Reviewed the full diff (backend filters, service layer, route tree, shared components, nav). This is a clean, faithful implementation of the issue's "reuse, don't fork" requirement. **Approving with one optional cosmetic nit.**

### What I verified

**Reuse mechanism ✅** — `core.ts` points the project route tree at the same five page modules as the workspace aggregate (with explicit route ids to avoid the file-derived id collision). Each page reads an optional `projectId` from the route and threads it into service calls, SWR keys, and link bases; the workspace routes pass nothing and keep today's behaviour. The contacts sidebar was extracted into one shared `RunnersSideNav` used by both layouts. No forked component copies.

**Backend filters ✅** — `runs`, `approvals`, `chat/sessions`, and `runners` each gained an optional `?project=<uuid>` narrowing to `pod__project_id`. All model FK paths (`Pod.project`, `Runner.pod`, `AgentRun.pod`, `AgentChatSession.pod`, `ApprovalRequest→agent_run→pod`) exist and are valid lookups.

**Private-runner rule preserved ✅** — on `runs.py` the project filter is a *separate chained* `.filter()`, so it AND-combines with the existing private-runner `Q` gate rather than widening it. `test_runs_project_filter_preserves_private_runner_rule` pins exactly this (a run on another user's private runner stays hidden even for a workspace admin under `?project=`). Pod-less / Cloud-Agent runs are intentionally excluded from a project view — consistent with the parent-issue scope split.

**Pagination scoping ✅** — `total_count`/`total_pages` are computed after the project filter, and `projectId` is part of the runs SWR key, so paging can't leak rows across scopes.

**Nav relocation ✅** — workspace `runners` entry filtered out of the top-level list via `RELOCATED_KEYS` and surfaced under "More" via `MORE_ITEM_KEYS`; new project "AI Workers" item added as a sibling of Schedulers, gated on project ADMIN/MEMBER.

**Access decision** — project panel gates on project ADMIN/MEMBER (`ProjectRunnersLayout`), mirroring the workspace panel's ADMIN/MEMBER rule at project scope. Reasonable resolution of the question the issue left open.

**Validation** — `pnpm turbo run check:types --filter=web --filter=@pi-dash/services` passes locally here (react-router typegen + tsc, no duplicate-id error).

### Optional nit (non-blocking, cosmetic)

`more-section.tsx`: the workspace-aggregate "AI Workers" item highlights via `pathname.includes("/runners/")` and `MORE_PATH_SEGMENTS` auto-opens on `/runners`. Because the project route is `/<ws>/projects/<pid>/runners/...`, visiting a project's AI Workers *sub*-routes (runs/approvals/chat/detail) also matches — so the workspace "More → AI Workers" row can auto-open the disclosure and show as active at the same time as the project's own "AI Workers" item. Purely visual; worth tightening the match to a workspace-root-anchored path if you want to avoid the double-highlight, but not required for this PR.

PR #348, opened 2026-09-04
PDASHOSS01-100 Add AgentRun.agent_metadata JSON field to durably record the local agent session id

irichard00:
## Summary

Adds a general-purpose `AgentRun.agent_metadata = models.JSONField(default=dict, blank=True)` column and uses it to durably record the local AI agent session id reported by the runner at run start.

AgentRun's four existing JSON columns are each an owned single-producer/consumer contract (`run_config`, `done_payload`, `prompt_manifest`, `required_capabilities`), so none is a safe home for general run metadata. `thread_id` holds the session id today but is a **resume handle**: it is cleared on retry (`run_lifecycle.apply_run_resume_unavailable`) and its meaning is agent-kind dependent. `agent_metadata` is **write-once-per-attempt and never cleared**, so a failed attempt's session id survives for audit / local-session lookup.

## Cloud (`apps/api`)

- `AgentRun.agent_metadata` field + migration `0024`.
- `RunStartedEndpoint` accepts `local_session_id` / `agent_kind` alongside the existing `thread_id` (write unchanged) and populates `agent_metadata` with `{local_session_id, local_thread_id, agent_kind}`. `local_thread_id` defaults to `thread_id` for older runners.
- Exposes `agent_metadata` on `AgentRunSerializer`.

## Runner (Rust)

- `AgentCursor::agent_kind()` maps each CLI to its wire string (`codex | claude_code | cursor_agent | openclaw | grok`).
- `ClientMsg::RunStarted` carries `local_session_id` + `agent_kind`; the supervisor aliases `local_session_id` to the thread id, mirroring the chat bridge's warm path. The genuinely per-kind distinct session id noted in the issue is left for a future design pass.

## Tests

- run-started persists the new keys, and stays backward compatible for a `thread_id`-only runner.
- a resume-unavailable retry that clears the parent's `thread_id` leaves its `agent_metadata` intact.

Validated: 21/21 in `test_https_transport_run_endpoints.py` pass; `cargo check` clean; `makemigrations --check` shows no missing migration for the new field.

## Open questions (from the issue)

- `agent_kind` kept in JSON rather than a typed column (keeps the migration to one field).
- No backfill (existing rows' `thread_id` only maps to a session id for Claude Code, and only where not already cleared — not worth it).

### DIFF
--- a/apps/api/pi_dash/runner/migrations/0024_agentrun_agent_metadata.py
+++ b/apps/api/pi_dash/runner/migrations/0024_agentrun_agent_metadata.py
@@ -1,0 +1,25 @@
+# Copyright (c) 2023-present Pi Dash Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
+"""Add ``AgentRun.agent_metadata`` — a general-purpose, write-once-per-attempt
+JSON column for run metadata that has no owned single-producer/consumer
+contract. Used to durably record the runner-reported local agent session id /
+thread id / agent kind at run start; unlike ``thread_id`` it is never cleared
+on retry, so a failed attempt's session id survives for audit."""
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("runner", "0023_one_active_includes_cancel_requested"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="agentrun",
+            name="agent_metadata",
+            field=models.JSONField(blank=True, default=dict),
+        ),
+    ]
--- a/apps/api/pi_dash/runner/models.py
+++ b/apps/api/pi_dash/runner/models.py
@@ -938,6 +938,16 @@ class AgentRun(models.Model):
     run_config = models.JSONField(default=dict, blank=True)
     required_capabilities = models.JSONField(default=list, blank=True)
     thread_id = models.CharField(max_length=128, blank=True, default="")
+    # General-purpose per-attempt run metadata. Unlike the four owned JSON
+    # contracts above (``run_config`` / ``done_payload`` / ``prompt_manifest``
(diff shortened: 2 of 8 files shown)


irichard00 (commented):
## Review — approved ✅

Reviewed the full diff against the issue's acceptance criteria. This is a tight, well-scoped change and I found no correctness issues.

**What holds up**
- **Migration** `0024_agentrun_agent_metadata` chains off `0023` (latest on `main`, no collision), `JSONField(default=dict, blank=True)` — non-nullable as required, and the model/migration field defs match (no `makemigrations` drift for this field).
- **Endpoint** — `RunStartedEndpoint` writes `agent_metadata` via `AgentRun.objects.filter(...).update(**updates)`, so the new key is actually persisted alongside the unchanged `thread_id` write. `local_thread_id` correctly falls back to `thread_id` for older runners; unreported keys land as `""` (not missing), matching the backward-compat criterion.
- **Retry survival** — the change deliberately leaves `run_lifecycle.apply_run_resume_unavailable` untouched, so clearing the parent's `thread_id` on a resume-unavailable requeue leaves `agent_metadata` intact. Good that this is pinned by a dedicated test rather than left implicit.
- **Runner** — `AgentCursor::agent_kind()` is an exhaustive match over all five variants (`codex | claude_code | cursor_agent | openclaw | grok`), so a future 6th agent kind fails to compile rather than silently emitting `""`. `local_session_id` aliases the thread id, mirroring the chat bridge's warm path — the genuinely per-kind distinct session id is correctly deferred per the issue's noted design pass.

**Notes (non-blocking, no action needed)**
- The runner never sends `local_thread_id`; the cloud always derives it from `thread_id`. That's consistent with the current alias design — flagging only so it's a conscious choice for the deferred design pass.
- Codex path stores `local_session_id == thread_id` today even though Codex's thread_id is a genuinely distinct concept from its session id; this is the known alias limitation the issue calls out, not a defect in this PR.

No changes requested.

PR #349, opened 2026-09-04
PDASHOSS01-101 Research Meta Muse Code as Pi Dash runner

irichard00:
## What

Implements Meta **Muse Code** as a first-class Pi Dash runner backend (`AgentKind::MuseCode`), driven in its headless one-shot mode `muse exec --json --prompt-file <PATH>`. Per the issue discussion, the research concluded Muse Code has no Codex-style app-server daemon but does expose a headless one-shot mode; the human confirmed "headless CLI mode is enough to start, let's implement the integration in this task directly."

Modeled on the existing **Cursor** adapter (one-shot per turn, JSONL on stdout). Two differences:
- The prompt is delivered via a temp `--prompt-file` (RAII-cleaned in the OS temp dir, no new dependency) instead of argv.
- MVP approval posture uses `--yolo` (mirrors Cursor's `--force`).

Muse Code is closed-source and its JSONL schema isn't fully documented, so the schema layer is deliberately tolerant: `type`/`subtype` dispatch with an `Unknown` catch-all, and run setup does **not** hard-depend on a `system/init` frame (any first frame starts the run; its `session_id` is adopted or synthesized).

## Runner changes

- New `runner/src/muse_code/` module (`schema` / `process` / `bridge` / `mod`) + `lib.rs` registration.
- `MuseCode` wired through every dispatch point: config schema (enum, `stall_timeout`, `display_name`, `default_binary` = `muse`, `install_page_url`, `MuseCodeSection`, `RunnerConfig.muse_code`), the `AgentBridge`/`AgentCursor` enums and all methods, CLI `runner add` (`agent_sections_for`, model applicability), `pidash doctor` (binary probe + `META_API_KEY` auth hint), supervisor crash classification, and the TUI config editor.

## Web changes

- `muse-code` added to `TRunnerAgent`, the model-option Records, the add-runner modal options + label, and an i18n label — selectable in the UI. The CLI path `pidash runner add --agent muse-code` also works.

## Testing

- New `runner/tests/muse_code_bridge_fake.rs`: **8 end-to-end tests** drive a fake `muse` CLI through init → assistant → tool_call → result, plus error results, no-init streaming, `warm`, drop-kills-child, and prompt shutdown. All pass.
- Unit tests for schema dispatch + command construction; all touched-module tests pass (`cargo test --lib` = 463 passed; the 2 failures are pre-existing, environment-specific, in files this PR does not touch).
- `cargo clippy --lib --tests` clean; web + i18n `check:types` clean.

## Notes / assumptions

- Cannot run the real `muse` binary in CI, so `muse exec`'s exact flag spellings (`--json`, `--prompt-file`, `--yolo`, `--session-id`, `--model`) and JSONL field names are taken from public Muse Code docs and treated as assumptions; the tolerant schema + fake-CLI harness bound the risk.
- Non-technical caveats flagged during research (Muse Code is closed-source; near-free access is contingent on Meta training on prompts/completions) are product decisions for the maintainers, not blockers to this adapter.

### DIFF
--- a/apps/web/core/components/runners/add-runner-modal.tsx
+++ b/apps/web/core/components/runners/add-runner-modal.tsx
@@ -35,7 +35,7 @@ type Props = {
 
 // Mirrors the runner CLI's ``--agent`` value-enum (kebab-case). Keep in
 // sync with runner/src/config/schema.rs:AgentKind.
-const AGENT_OPTIONS = ["claude-code", "codex", "cursor-agent", "open-claw", "grok"] as const;
+const AGENT_OPTIONS = ["claude-code", "codex", "cursor-agent", "open-claw", "grok", "muse-code"] as const;
 type TAgent = (typeof AGENT_OPTIONS)[number];
 const DEFAULT_AGENT: TAgent = "claude-code";
 const RUNNER_NAME_WHITESPACE_RE = /\s/;
@@ -227,6 +227,7 @@ export const AddRunnerModal = observer(function AddRunnerModal(props: Props) {
     if (value === "cursor-agent") return t("Cursor");
     if (value === "open-claw") return t("OpenClaw");
     if (value === "grok") return t("Grok");
+    if (value === "muse-code") return t("Muse Code");
     return t("Codex");
   };
 
--- a/apps/web/core/components/runners/runner-models.ts
+++ b/apps/web/core/components/runners/runner-models.ts
@@ -18,7 +18,7 @@
  */
 
 // Mirrors the runner CLI's ``--agent`` value-enum (kebab-case).
-export type TRunnerAgent = "claude-code" | "codex" | "cursor-agent" | "open-claw" | "grok";
+export type TRunnerAgent = "claude-code" | "codex" | "cursor-agent" | "open-claw" | "grok" | "muse-code";
 
 export interface IRunnerModelOption {
   /** Unique select value within an agent's list. */
@@ -116,6 +116,9 @@ export const RUNNER_MODEL_OPTIONS: Record<TRunnerAgent, IRunnerModelOption[]> = 
   "cursor-agent": [DEFAULT_OPTION, ...CURSOR_OPTIONS],
   "open-claw": [DEFAULT_OPTION],
   grok: [DEFAULT_OPTION, ...GROK_OPTIONS],
+  // Muse Code's model slug space is provider-specific (Meta Model API); the
+  // runner accepts any non-empty slug, so offer just the agent default here.
+  "muse-code": [DEFAULT_OPTION],
 };
 
 /**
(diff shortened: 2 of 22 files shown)


irichard00:
## Review — Muse Code runner backend (PDASHOSS01-101)

Reviewed the full diff and independently re-ran the validation on branch `pi-dash/pdashoss01-101` (`b1ee62fd`). **Approving.** The adapter is clean, tolerant-by-design, and wired through every dispatch point; I found no correctness defects.

### Independently verified
- `cargo test --test muse_code_bridge_fake` → **8 passed**
- `cargo test --lib muse_code` → **9 passed**
- `cargo build` → clean · `cargo clippy --lib --tests` → clean
- `pnpm --filter=web check:types` and `--filter=@pi-dash/i18n check:types` → clean
- Wiring spot-checked as exhaustive: `AgentKind::MuseCode` in `config/schema.rs` (stall timeout, display name, default binary `muse`, install URL, `MuseCodeSection`), `agent/mod.rs` (bridge/cursor arms), `cli/doctor.rs` (blocking `muse` probe + non-blocking `muse-auth`/`META_API_KEY` hint), `cli/runner_ops.rs`, `daemon/supervisor.rs` crash classification, TUI config, and the web add-runner modal + i18n label.

### Design notes (verified sound, no change requested)
- **Tolerant schema** — `type`-tagged dispatch with an `Unknown` catch-all, and `read_events` falls back to `Unknown(Value)` on unparsable JSON, so upstream schema drift degrades to "preserved but unmapped" rather than crashing the daemon. Good.
- **No hard `system/init` dependency** — `wait_for_init` adopts any first frame's `session_id` (or synthesizes `muse-<run_id>`) and buffers a non-init first frame so a terminal-error-first stream still surfaces via the normal pump. Correct.
- **Prompt-file lifetime** — `PromptFile` is created in the OS temp dir (never the worktree) and unlinked on drop via the `MuseProcess` guard. No leftover files, no git-status pollution.
- **EOF without a terminal frame** — if `muse exec` exits without a `result` frame, `next_events` returns `None` and the supervisor reports `RunFailed{"agent stdout closed"}` (supervisor.rs:3167), matching every other one-shot bridge. Acceptable.

### The one caveat (already flagged — not a blocker)
The exact `muse exec` flag spellings (`--json`, `--prompt-file`, `--yolo`, `--session-id`, `--model`) and the JSONL field names are assumptions from public docs; the real (closed-source) binary is unavailable in this environment, so they can't be verified in-session. The tolerant schema bounds the blast radius, but two behaviours specifically depend on unverified flag *semantics* and should be watched on the first real run:
- `--session-id <id>` on `exec` must **resume** an existing session (not create/reject) for cross-turn continuity to work.
- `--yolo` must be the correct spelling for the auto-approval/sandbox-off posture (MVP, mirrors Cursor's `--force`; approvals aren't wired and `send_approval` fails loud by design).

Recommend confirming the event mapping and flag surface against the live Meta `muse` CLI before enabling this backend for production runs.

### Nit (optional)
`bridge.rs:204-205` — `deadline` is computed and then `remaining` is immediately derived from it against `Instant::now()`, which just reproduces `INIT_TIMEOUT`. Harmless; could be simplified to pass `INIT_TIMEOUT` directly to the `timeout`.

PR #350, opened 2026-09-04
PDASHOSS01-103 Add delete option for personal dev machine connections

irichard00:
## Summary

The **AI Dev Machines** page (`/:workspaceSlug/ai-dev-machines`) listed each connected dev machine with **Rotate** and **Revoke** actions but no way to remove the connection entirely. This adds a **Delete** action — the destructive counterpart to Revoke (which soft-marks the machine revoked and keeps its history in the list).

## Changes

**Backend**
- `delete_dev_machine` service (`runner/services/runner_delete.py`): invalidates the machine's tokens, tears down its runners (enqueue daemon frame → revoke → close → delete row, mirroring `delete_runner`'s ordering so the offline buffer doesn't strand a frame), then hard-deletes the machine row (cascading its `MachineSession` rows).
- `DevMachineDeleteEndpoint` — `DELETE /api/runners/dev-machines/<id>/`, owner-scoped like the existing revoke/rotate actions, honoring `?purge_local=true|false` (default `true`).

**Frontend**
- `RunnerService.deleteDevMachine` + a Delete button and confirm modal on the page (available for revoked machines too).
- i18n keys.

## Personal, not admin-only

Dev-machine add/list/revoke/rotate — and now delete — are all owner-scoped (`Visibility.PRIVATE`, `owner=request.user`) and the page gates ADMIN **and** MEMBER, so these actions are already personal, not admin-only. Delete follows the same rule.

## Testing

- 5 new API tests in `test_runner_delete_views.py` (default teardown, purge_local=false, non-owner 404, missing workspace 400, invalid flag 400). Full `pi_dash/tests/unit/runner/` suite: 408 passed.
- `pnpm check:types` and `check:lint` pass for web / services / i18n.

### DIFF
--- a/apps/api/pi_dash/runner/services/runner_delete.py
+++ b/apps/api/pi_dash/runner/services/runner_delete.py
@@ -34,8 +34,9 @@ from __future__ import annotations
 from uuid import UUID
 
 from django.db import transaction
+from django.utils import timezone
 
-from pi_dash.runner.models import Runner
+from pi_dash.runner.models import DevMachine, MachineToken, Runner
 from pi_dash.runner.services.pubsub import (
     close_runner_session,
     send_runner_remove,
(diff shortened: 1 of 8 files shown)


irichard00 (commented):
## Review — PDASHOSS01-103 Add delete option for personal dev machine connections

Reviewed commit `ef47fbb6` (8 files, +291/−17). **Approving** — the change is correct, closely mirrors the established `delete_runner` / `DevMachineRevokeEndpoint` patterns, and the acceptance criteria are all covered. A few non-blocking notes below.

### What I verified
- **Delete visibility** — the Delete button is rendered outside the `!machine.revoked_at` block, so it shows for both active and already-revoked machines while Rotate/Revoke stay hidden once revoked. Matches the AC.
- **Teardown ordering** — `delete_dev_machine` revokes tokens, enqueues the daemon control frame for every active runner *before* `runner.revoke()` closes the session (batched enqueue-then-revoke), then explicitly deletes the runner rows (the FK is `SET_NULL`, so a bare machine delete would orphan them) and finally the machine row. This mirrors `delete_runner`'s reasoning and avoids the enqueue-vs-evict window.
- **`purge_local` semantics** — `True` → `remove_runner` frame + canonical `runner_removed` reason (daemon wipes local state); `False` → `revoke` + non-canonical `user_revoke` (local install untouched). Identical to `delete_runner`.
- **Guard ordering** — missing `workspace` → 400, non-member → 403, invalid `purge_local` → 400 (before any mutation, machine untouched), out-of-scope machine → 404, success → 204. Owner scoping reuses `_machine_is_in_workspace_scope`.
- **Tests** — `test_dev_machine_delete_*` cover default teardown, `purge_local=false`, 404 other-user, 400 missing-workspace, and 400 invalid-purge. Solid coverage.

### Non-blocking notes
1. **Cross-workspace teardown.** `DevMachine` is user-scoped, not workspace-scoped (per the model docstring: one laptop can host runners for multiple workspaces). `delete_dev_machine` tears down *all* of the machine's runners and tokens (`filter(dev_machine_id=...)`, unscoped by workspace) and deletes the machine row — so a user deleting the machine from workspace W1's page also removes its runners/tokens in their W2. This is exactly what the existing `DevMachineRevokeEndpoint` already does, so it's consistent with the user-scoped model; flagging only to confirm it's the intended semantics for delete too.
2. **Test gaps (minor).** No explicit 403 non-member test for the delete endpoint (the `is_workspace_member` guard is shared with revoke/rotate, which are tested), and no web-component test for the new modal (the page has no test harness — consistent with the existing untested modal pattern). Both were called out in the handoff.
3. **Transaction scope (negligible).** Unlike revoke, the delete endpoint does the scope check + machine fetch outside the service transaction and doesn't `select_for_update` the machine row. Concurrent deletes are idempotent (`filter().delete()` on an already-deleted row is a no-op; frames are advisory), so this is harmless.

No changes required from my side.

PR #351, opened 2026-09-04
PDASHOSS01-104 Add Pi Dash core concept video to onboarding first slide

irichard00:
## What

Adds the Pi Dash core concept video ([youtu.be/-sitbwrwjno](https://youtu.be/-sitbwrwjno)) to the first-sign-in onboarding.

The onboarding tour (`TourRoot`, shown fullscreen on the home page while `is_tour_completed` is false) opens on a **welcome** screen. This replaces that screen's static PiSymbol hero with a responsive YouTube embed and updates the heading/copy to invite new users to watch the video before taking the product tour or diving in.

## Why

Per PDASHOSS01-104 — surface the core concept video as the first thing new users see during onboarding.

## Changes

- `apps/web/ce/components/onboarding/tour/root.tsx`:
  - Add `CORE_CONCEPT_VIDEO_EMBED_URL` constant.
  - Replace the welcome-screen PiSymbol hero with a 16:9 (`aspect-video`) sandboxed YouTube `<iframe>`.
  - Update the welcome heading to "Welcome to Pi Dash{, name} — watch the core concept video" and refresh the supporting copy.
  - Drop the now-unused `PiSymbol` import.

## Design note

The issue said "make it the first slide of the onboarding page". The welcome screen is literally the first screen every new user sees, and placing the video there (rather than a deeper carousel slide gated behind the "Take a Product Tour" button) guarantees the video is visible immediately. Happy to instead add it as a dedicated navigable carousel slide if that's preferred.

## Testing

- `oxlint` on the changed file: 0 warnings / 0 errors.
- `pnpm --filter=web check:types`: passes.
- The `sandbox` attribute intentionally omits `allow-same-origin` (the lint rule forbids combining it with `allow-scripts`); a public YouTube embed plays without it.

### DIFF
--- a/apps/web/ce/components/onboarding/tour/root.tsx
+++ b/apps/web/ce/components/onboarding/tour/root.tsx
@@ -8,7 +8,7 @@ import { useState } from "react";
 import { observer } from "mobx-react";
 // pi dash imports
 import { Button } from "@pi-dash/propel/button";
-import { CloseIcon, PiSymbol } from "@pi-dash/propel/icons";
+import { CloseIcon } from "@pi-dash/propel/icons";
 // assets
 // TODO(onboarding): replace these reused Plane-era screenshots with Pi Dash-native captures
 // (runner CLI, runner chat, workpad) once design provides them.
@@ -28,6 +28,9 @@ export type TOnboardingTourProps = {
 };
 
 export type TTourSteps = "welcome" | "work-items" | "runners" | "chat" | "workpads" | "pages";
+
+// Pi Dash core concept video shown on the welcome screen (https://youtu.be/-sitbwrwjno).
+const CORE_CONCEPT_VIDEO_EMBED_URL = "https://www.youtube.com/embed/-sitbwrwjno";
 
 const TOUR_STEPS: {
   key: TTourSteps;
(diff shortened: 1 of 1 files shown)


irichard00:
## Code review — PDASHOSS01-104 (onboarding core-concept video)

Reviewed the diff on `apps/web/ce/components/onboarding/tour/root.tsx`. The change is clean and matches the issue at the code level: the welcome screen now embeds the core-concept video (`-sitbwrwjno`) in a 16:9 (`aspect-video`) hero, the heading invites watching it, and the "Take a Product Tour" / "No thanks…" paths and the five carousel slides are untouched. `PiSymbol` import correctly dropped.

### One risk to verify in a browser before this ships: the sandbox may block playback

The iframe uses `sandbox="allow-scripts allow-presentation allow-popups allow-popups-to-escape-sandbox"` — deliberately **without** `allow-same-origin`. Omitting it forces the framed document into an opaque origin, and the YouTube player frequently fails under that (it needs same-origin access to its own cookies/storage), showing an error instead of playing. Autoplay isn't set, so this only surfaces on click. **The core acceptance criterion ("video plays inline when clicked") can't be confirmed from static review — it needs a real browser check in the In Test phase.**

I verified *why* `allow-same-origin` was omitted, and the constraint is real — the `react/iframe-missing-sandbox` rule is a catch-22 here (oxlint 1.51.0, repo config, `--max-warnings=11957` gate is pinned so any net-new warning fails CI):

| sandbox | oxlint result |
|---|---|
| *(no sandbox attr)* | 1 warning — rule requires a sandbox → fails gate |
| `allow-scripts allow-same-origin …` | 1 warning — "Remove allow-scripts or allow-same-origin" → fails gate |
| `allow-scripts …` (no same-origin) — **this PR** | **0 warnings** ✓ |

So the PR's config is the only lint-clean sandbox. The rule's "dangerous combination" warning is actually a **false positive** for a cross-origin YouTube embed (the escape-the-sandbox attack it guards against requires the framed content be same-origin as the app; youtube.com is not).

### Fallback if playback fails in test

Suppress the false-positive on that one line and add `allow-same-origin` (the standard, known-good config for a sandboxed YouTube embed):

```tsx
// react/iframe-missing-sandbox false-positive: cross-origin YouTube embed cannot
// escape into the app origin, so allow-scripts + allow-same-origin is safe here.
// eslint-disable-next-line react/iframe-missing-sandbox
<iframe
  src={CORE_CONCEPT_VIDEO_EMBED_URL}
  …
  sandbox="allow-scripts allow-same-origin allow-presentation allow-popups allow-popups-to-escape-sandbox"
  allowFullScreen
/>
```

(oxlint honors `// oxlint-disable-next-line react/iframe-missing-sandbox` too; confirm which directive form this repo suppresses with before committing.)

### Nits (optional, non-blocking)

- Consider `loading="lazy"` on the iframe so YouTube isn't fetched until the onboarding modal actually mounts.

No changes pushed — the code is correct as written and the only open item is a runtime check that requires a browser. Recommend In Test verify playback specifically, and apply the fallback above only if it fails.

irichard00:
Pushed f6db8670: added `allow-same-origin` to the onboarding iframe sandbox and suppressed the `react/iframe-missing-sandbox` false positive on that attribute (YouTube is cross-origin, so scripts + same-origin cannot escape the sandbox). Re-verified in headless Chrome over http://localhost: old markup = blank frame; new markup renders the player and plays (autoplay+mute probe shows live frames). oxlint 0 warnings, oxfmt clean, no new tsc errors in the file.

irichard00:
Review pass on `f6db8670`: approved.

- `sandbox` now includes `allow-same-origin`. That's the fix test-verified earlier (player renders and plays). The `oxlint-disable-next-line react/iframe-missing-sandbox` sits directly above the `sandbox=` attribute, which is where oxlint reports the warning, and the comment explains why it's a false positive (YouTube is cross-origin).
- `oxlint --deny-warnings` on the branch's `root.tsx`: 0 warnings, 0 errors.
- Nothing else changed in this commit. `TOUR_STEPS`, the tour and dismiss handlers, and the heading and copy are the same as in the first commit.

No findings. Moving to In Test so the flow can be exercised in the app.

PR #355, opened 2026-09-07
PDASHOSS01-125 Windows MSI: launch pidash auth login from the installer ExitDialog instead of asking the user to run it manually

irichard00:
## Problem

Every install path except the Windows MSI auto-starts the device-code login (`install.sh`, `install.ps1`, and bare `pidash`). The MSI just dropped `pidash.exe` and exited, forcing the user to open a terminal and type `pidash` themselves. An interactive double-click install should finish setup in the same sitting.

## Change

- **`runner/wix/setup.cmd`** (new) — a shim that runs `pidash auth login` then `pause`s. `pidash.exe` is a console-subsystem app; ShellExecute-ing it directly from the GUI installer would allocate a console that vanishes the instant the process exits, so a failed login would show nothing. The shim keeps the window open.
- **`runner/wix/main.wxs`**:
  - `SetupCmd` component/`File` installed alongside the binary in the `Bin` dir.
  - `WIXUI_EXITDIALOGOPTIONALCHECKBOX` + `...TEXT` ("Sign in to Pi Dash now"), rendered by `WixUI_FeatureTree`'s ExitDialog, ticked by default.
  - `LaunchPidashSetup` — a **type-18** custom action (`FileKey` + `ExeCommand`), chosen so no `WixUtilExtension` / extra `light.exe -ext` flag needs threading through the cargo-dist build. `Impersonate='yes'` runs it **unelevated as the installing user** so the token + workspace binding land under that user's profile (`login.rs:148-152`), not SYSTEM's. `Return='asyncNoWait'` lets the paused console outlive `msiexec`.
  - ExitDialog `Publish` gated on `WIXUI_EXITDIALOGOPTIONALCHECKBOX = 1 and NOT Installed`.
  - "Sign in to Pi Dash" Start Menu shortcut (the `README.md:81` follow-up) — the fallback for silent installs where nothing can be popped.
- **`runner/README.md`** — MSI section rewritten to describe the checkbox + shortcut.

Silent installs (`msiexec /qn`, `/qb`) skip the UI sequence entirely, so nothing spawns; repairs/upgrades are excluded via `NOT Installed`.

## Verification status

XML well-formedness verified locally (`xmllint`). **A full WiX build + manual QA needs a Windows x64 host and the WiX toolchain — not available in the authoring environment** — so this is the In Test phase's job. Open questions to confirm on a real build:

1. Type-18 CA (`FileKey`→`.cmd`, `ExeCommand=""`) actually launches the shim. `CustomAction` type 18 uses CreateProcess semantics, which cannot exec a `.cmd` directly; if the CA no-ops on a real build, the fallback is a `WixShellExec` CA (adds `WixUtilExtension`) or converting the shim to a real `.exe`.
2. `Return="asyncNoWait"` + `pause` — the paused console must survive `msiexec` exiting.
3. `/qb` (basic UI) skips the ExitDialog like `/qn` does.

### DIFF
--- a/runner/README.md
+++ b/runner/README.md
@@ -71,14 +71,18 @@ Or on Windows, download and run the MSI installer:
 
 <https://github.com/The-AI-Republic/pi-dash/releases/latest/download/pidash-x86_64-pc-windows-msvc.msi>
 
-The MSI is a standard double-click Windows Installer — it drops `pidash.exe` and exits. It does **not** automatically launch the auth flow (MSI typically runs elevated as the Windows Installer service, and many MSI invocations are silent for IT/Group Policy deploys, so spawning a terminal from it would be wrong). To finish setup after the MSI closes:
+The MSI is a standard double-click Windows Installer — it drops `pidash.exe`, adds a **Sign in to Pi Dash** Start Menu shortcut, and (on an interactive install) offers to finish setup for you. The final installer screen shows a **Sign in to Pi Dash now** checkbox, ticked by default; leaving it ticked and clicking **Finish** opens a console and runs `pidash auth login`, walking you through the same device-code flow as the `install.ps1` one-liner.
+
+The auto-launch is deliberately scoped to interactive installs. Silent invocations (`msiexec /qn`, and the basic-UI `/qb`) — the IT/Group Policy/Intune/SCCM deploy paths — skip the installer UI entirely, so no terminal is ever spawned under the Windows Installer service. For those deploys, the admin ships the MSI and the user completes sign-in later from the **Sign in to Pi Dash** Start Menu shortcut (or by opening PowerShell and running `pidash`). Repairs and upgrades don't re-launch it either.
+
+If you skipped the checkbox, you can finish setup any time from the Start Menu shortcut, or:
 
 ```powershell
 # Open PowerShell and run:
 pidash
 ```
 
-With no config yet, bare `pidash` drops into `auth login` and walks you through the same device-code flow as the `install.ps1` one-liner. (A discoverable Start Menu shortcut for this step is tracked as a follow-up.)
+With no config yet, bare `pidash` drops into `auth login` and walks you through the device-code flow.
 
 Windows release assets also include a `pidash-x86_64-pc-windows-msvc.zip` archive with `pidash.exe` for advanced/manual installs.
 
--- a/runner/wix/main.wxs
+++ b/runner/wix/main.wxs
@@ -128,7 +128,46 @@
                                 Source='$(var.CargoTargetBinDir)\pidash.exe'
                                 KeyPath='yes'/>
                         </Component>
+
+                        <!--
+                          Post-install setup shim. Launched (a) from the ExitDialog
+                          checkbox on interactive installs and (b) from the Start Menu
+                          shortcut. Runs `pidash auth login` and pauses so the console
+                          survives the login flow. See wix\setup.cmd for the rationale.
+                        -->
+                        <Component Id='SetupCmd' Guid='*'>
+                            <File
+                                Id='setupCmd'
+                                Name='setup.cmd'
+                                DiskId='1'
+                                Source='wix\setup.cmd'
+                                KeyPath='yes'/>
+                        </Component>
                     </Directory>
+                </Directory>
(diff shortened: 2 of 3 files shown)


irichard00:
## Review — 1 confirmed bug, fixed in this pass

**Confirmed correctness bug (fixed): the `LaunchPidashSetup` custom action could never launch the shim.**

As committed (`91c10e9b`), the custom action was **type-18** (`FileKey='setupCmd' ExeCommand=''`). A type-18 CA launches its target file via `CreateProcess`, which **cannot execute a `.cmd`/`.bat`** — a batch file is not a PE image, so the call fails with error 193. The "Sign in to Pi Dash now" checkbox would tick, the user would click **Finish**, and nothing (or an error) would happen. This is exactly the failure the issue description flagged as "the single most likely thing to need a follow-up" — but it's a documented Windows certainty, not a genuine unknown, so it's worth fixing before the change reaches Windows QA rather than spending a QA cycle rediscovering it.

**Fix (`d03e07d9`):** switch to a **type-34** CA (`Directory` + `ExeCommand`) that invokes `cmd.exe` explicitly as the batch host:

```
[System64Folder]cmd.exe /c start "Pi Dash setup" "[#setupCmd]"
```

- `start` gives the shim its own console window, not tied to msiexec's lifetime; with `Return='asyncNoWait'` the paused console outlives the installer.
- `Impersonate='yes'` is retained, so the device-code token still lands under the installing user's profile, not SYSTEM's.
- Needs no `WixUtilExtension`, so nothing extra is threaded through the cargo-dist build.
- Also declares `<Directory Id='System64Folder'/>` so `[System64Folder]` resolves rather than expanding empty.

The **Start Menu shortcut is unaffected** — Explorer/Start launch `.cmd` targets via `ShellExecute`, which honors the file association and runs `cmd.exe`, so that path was already correct.

### Verified from Linux
- XML well-formedness (`xmllint --noout`) → OK.
- All Id references resolve (`Bin`, `System64Folder`, `setupCmd`); type-34 attribute combination is the documented WiX pattern for launching a batch file from a CA.
- (Full WiX XSD validation isn't possible here — the WiX v3 schema's .NET regex facets don't compile under libxml2. Not a defect in this file.)

### Still requires Windows x64 build + QA (In Test)
The full matrix in the test-handoff comment stands — checkbox ticked → console + device-code flow; `config.toml` under the user's profile; unticked → nothing; `/qn` + `/qb` → nothing spawns, Start Menu shortcut works; repair/upgrade → no spawn (`NOT Installed`); console survives msiexec exit; and that `light.exe` builds the package clean (ICE/GUID).

_Note: an earlier issue comment described this fix (type-34 + a `pidash-setup.cmd` rename) as already implemented, but it was never committed — the repo still carried the broken type-18 CA and `setup.cmd`. This pass actually applies the type-34 fix to `setup.cmd`._

irichard00:
Closing in favour of #356, which solves the same issue and is further along.

Both PRs were opened for PDASHOSS01-125 — an agent picked the issue up minutes after it was filed, and I worked on it in parallel without noticing. My fault for not checking.

Keeping #356 because it is strictly ahead on the same problems:

- **`cargo dist plan` fails here.** This branch edits `runner/wix/main.wxs` without adding `"msi"` to `allow-dirty` in `dist-workspace.toml`, so cargo-dist rejects the hand-customized template. `plan` is the release workflow's first job, so a release would abort before building anything. #356 fixes this.
- **`Source='wix\\setup.cmd'`** — verified failing in a real build (`pidash-v0.1.23-rc.1`, `light` exit 103, file not found). The path resolves against cargo-dist's cwd, not the package root. #356 uses `$(sys.SOURCEFILEDIR)`, which fixed it.
- **No `WIX_UPGRADE_DETECTED` guard**, so a major upgrade re-runs the full device-code login on an already-authenticated machine.
- `Guid='*'` on a component whose keypath is a registry value rather than a file.

Credit where due: this PR independently reached the same type-34 custom action conclusion, for the same reason (`CreateProcess` cannot launch a `.cmd`). That agreement is decent evidence the approach is right.

One thing still unsolved in both: the MSI fails ICE validation (`light` exit 204), most likely the Start Menu shortcut's HKLM keypath under ICE43. Tracked on #356.

PR #361, opened 2026-09-08
PDASHOSS01-130 Dev machine status should reflect the machine's own connection, not its runner count

irichard00:
## Problem

On `/:workspaceSlug/ai-dev-machines/`, a dev machine's status badge was derived **entirely from runner counts**. A machine with a live, connected daemon but zero runners showed **"Registered"** — visually identical to a machine enrolled long ago that has never connected. The status is supposed to describe the *machine* (is the daemon reachable right now?), not its runners — runner health already has its own column.

## Root cause

`getDevMachineStatus` (`apps/web/app/(all)/[workspaceSlug]/ai-dev-machines/page.tsx`) branched on `online_runner_count` / `runner_count`, so a runner-less machine always fell through to `registered`. The connection signal (`control_online`, an `Exists()` over recently-seen `MachineSession` rows) was already computed and serialized on `DevMachineSerializer` but never read by the frontend.

## Change

Frontend-only. Status now reflects the machine's own connection:

- `revoked_at` → **Revoked**
- `control_online` → **Connected** (daemon reachable now)
- `last_seen_at` → **Offline** (seen before, not connected now)
- otherwise → **Registered** (enrolled, never connected)

The new **Connected** status replaces the runner-count-based **Active** machine status, so the badge no longer overloads the meaning of "Active" — runner health stays in the existing active/total runner column. Adds the `Connected` string to the `en` locale (the only locale under `packages/i18n/src/locales/`).

## Validation

- `oxlint` clean on the changed page.
- `pnpm run check:types`: no new errors in the touched files (remaining errors are pre-existing in unrelated files).

### DIFF
--- a/apps/web/app/(all)/[workspaceSlug]/ai-dev-machines/page.tsx
+++ b/apps/web/app/(all)/[workspaceSlug]/ai-dev-machines/page.tsx
@@ -35,17 +35,22 @@ Start-Process msiexec.exe -Wait -ArgumentList "/i \`"$msi\`""`;
 
 const service = new RunnerService();
 
-type DevMachineStatus = "active" | "offline" | "registered" | "revoked";
+// Status describes the *machine's* own connection, not its runners. Runner
+// health lives in the separate active/total runner column. "Connected" means
+// the daemon holds a recently-seen machine control session (``control_online``)
+// and can execute cloud-pushed commands right now — independent of whether any
+// runners are registered.
+type DevMachineStatus = "connected" | "offline" | "registered" | "revoked";
 
 const DEV_MACHINE_STATUS_BADGE_VARIANT: Record<DevMachineStatus, TBadgeVariant> = {
-  active: "accent-success",
+  connected: "accent-success",
   offline: "accent-neutral",
   registered: "accent-primary",
   revoked: "accent-warning",
 };
 
 const DEV_MACHINE_STATUS_I18N_LABELS: Record<DevMachineStatus, string> = {
-  active: "Active",
+  connected: "Connected",
   offline: "Offline",
   registered: "Registered",
   revoked: "Revoked",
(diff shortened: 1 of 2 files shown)


irichard00:
## Code review — approved

Reviewed the diff on `pi-dash/pdashoss01-130`. The change does exactly what the ticket asks, and I could not find a correctness, type, or i18n gap.

**Core logic** (`ai-dev-machines/page.tsx`) — `getDevMachineStatus` now branches on the machine's own connection instead of runner counts:
- `revoked_at` → `revoked`
- `control_online` → `connected` (new, `accent-success`)
- `last_seen_at` → `offline`
- else → `registered`

This maps 1:1 to the acceptance criteria, including the runner-less connected case that was the whole point of the ticket.

**Verified**
- `IDevMachine` (`packages/types/src/runner.ts`) carries `control_online: boolean`, `last_seen_at`, `revoked_at` — no API/model change needed, as the ticket said.
- `DevMachineStatus` union, `DEV_MACHINE_STATUS_BADGE_VARIANT`, and `DEV_MACHINE_STATUS_I18N_LABELS` all drop the old `active` key and add `connected` consistently — no dangling key, no overloaded "Active" meaning.
- Label renders via `t(DEV_MACHINE_STATUS_I18N_LABELS[status])`; the `Connected` key is present in `packages/i18n/src/locales/en/translations.ts` (the only locale in the repo). `Offline`/`Registered`/`Revoked` keys already exist.
- Runners column (`{active} active / {total} total`) is unchanged — machine status is now independent of runner counts.
- `getDevMachineStatus`/`DevMachineStatus` are not referenced anywhere else, so no other view needed updating.

The remaining hunks are pure Prettier reflow of existing `t(...)` JSX strings — no behavioral change.

**Not covered (test-phase, not a code defect):** no live UI run against a real connected daemon was possible in the coding session, and there's no unit test (the page has no test harness and `getDevMachineStatus` isn't exported). The In Test pass should confirm the connected-daemon-zero-runners case against a real or seeded machine.

PR #366, opened 2026-09-09
PDASHOSS01-131 Daemon never reports its AgentKind, so the cloud cannot identify or gate a runners agent

irichard00:
## Problem

The cloud had no idea which agent CLI a given runner drives. `Runner.capabilities` is the field shaped for it, but nothing ever wrote it — so failure diagnostics (`infer_agent_label`) had to guess the agent from the model slug, the runner name, the host label, and finally the raw error text. That guessing is unreliable in both directions (a Grok/Muse Code runner on its default model with a neutral host name renders the generic auth message; an ambiguous model slug can name the wrong vendor and send an operator to re-authenticate the wrong account).

## Fix (issue Consequence 1)

Have the daemon report, in the session-open (Hello) payload, the `AgentKind` the runner is configured with, and persist it as an exact signal:

- **Runner (Rust):** add `agent_kind` to `AttachBody` (`runner/src/cloud/http.rs`), populated from `inst.config.agent.kind` in `attach_body_for_instance` (`runner/src/daemon/supervisor.rs`). Serde renders it snake_case (`claude_code`, `muse_code`) — the spelling the diagnostics matcher already expects.
- **API (Python):** `apply_hello` (`session_service.py`) persists it as an `agent:<kind>` capability, which `infer_agent_label` already reads. An older daemon that omits the field leaves any existing capability untouched (no clobber); a malformed value is rejected by a strict charset check.
- Updated the pinned `test_capabilities_is_not_yet_populated_in_production` (the gap is now closed) and added `apply_hello` coverage for the populate / no-clobber / reject-malformed paths, plus a Rust test locking the snake_case wire spelling.

## Deferred (issue Consequence 2)

Gating the "Add runner" modal on the kinds a daemon supports (with an "upgrade this runner" hint instead of a raw parse error) is a larger, mostly-frontend change on a *different* transport — the dev-machine machine-control path, not the runner Hello — plus cloud persistence and web UI work. Recommend tracking it as a separate follow-up issue; flagged on the Pi Dash issue for triage.

## Testing

- `cargo build` + `cargo test` (runner): AttachBody key-set + snake_case spelling tests pass; the one unrelated `workspace::git::tests` failure is pre-existing on `main`.
- `pytest` (api): `test_run_diagnostics.py` (25) and `test_serializers_and_model.py` (16) pass, plus session-open view tests.

### DIFF
--- a/apps/api/pi_dash/runner/services/session_service.py
+++ b/apps/api/pi_dash/runner/services/session_service.py
@@ -12,6 +12,7 @@ Carved out of ``consumers.py`` per ``.ai_design/move_to_https/tasks.md``
 from __future__ import annotations
 
 import logging
+import re
 from datetime import datetime, timedelta
 from typing import Any, Dict, Optional
 from uuid import UUID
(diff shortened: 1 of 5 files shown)


irichard00:
## Review — approve (Consequence 1)

Reviewed the full diff and verified the wire contract end-to-end, live on this branch.

**Correctness — verified**
- All six `AgentKind` serde spellings line up exactly with `_match_agent_label`: `codex`→Codex, `claude_code`→Claude Code, `cursor_agent`→Cursor, `open_claw`→OpenClaw, `grok`→Grok (`\bgrok\b` matches at the `:` boundary in `agent:grok`), `muse_code`→Muse Code. `test_infer_agent_label_reads_agent_capability_when_present` pins every one.
- `AttachBody` is `#[derive(Serialize)]` only (outgoing-only), so the new **required** `agent_kind` field carries no deserialize/compat risk; a cloud that predates it just ignores the extra key.
- `_agent_capabilities` no-clobber + strict-charset logic is right: a missing field leaves an existing capability intact, a malformed value is rejected, and a legitimate change (kind A → kind B) updates.

**Acceptance criteria** — all six met (daemon reports snake_case kind; `apply_hello` persists `agent:<kind>`; `infer_agent_label` reads it; older-daemon no-clobber; malformed rejected; `capabilities` stays serializer-read-only, written server-side).

**Tests run here (not just trusted from the writeup)**
- Python: `test_run_diagnostics.py` + `test_serializers_and_model.py` → **41 passed**.
- Rust: `attach_body_reports_agent_kind_snake_case`, `attach_body_serialization_unchanged_by_observability_design` → **pass**.

**One nit — fixed in this review (pushed)**
The docstring on `test_infer_agent_label_reads_agent_capability_when_present` still said the capability path was "not yet exercised in production — see the test below," but this PR makes `apply_hello` write it and renamed the referenced test. Repointed it at `test_populated_capability_identifies_the_agent_over_neutral_signals`.

**Scope note (not blocking):** Consequence 2 (the "Add runner" modal gating on daemon-supported kinds) is correctly deferred — the AC does not require it and it lives on a different transport. Recommend filing the follow-up issue so it isn't lost.

LGTM.

PR #367, opened 2026-09-09
PDASHOSS01-132 No CI runs on push to main — two green PRs can merge into a main that does not build

irichard00:
## Problem

Nothing validated `main` after a merge. Every test workflow was `pull_request`-only, and the only `push:` triggers watched the wrong branches — `codeql.yml` on `preview`/`canary`/`master`, `release.yml` on tags. GitHub re-runs PR checks on `synchronize` (new commits), **not** when the base branch moves, so two PRs each green against their own older base can merge into a `main` that does not build.

This already happened on 2026-09-08: the runner crate failed to compile on `main` after #354 and #349 merged cleanly but produced a non-exhaustive `AgentKind` match. Fixed in #362, but the gap that let it through stayed open.

Separately, `pull-request-build-lint-api.yml` triggered only on PRs into `preview`, so ruff never gated any PR into `main`.

## Change

- Add a `push:` trigger on `main` to the three test workflows — `pull-request-test-runner.yml`, `pull-request-test-api.yml`, `pull-request-test-web-apps.yml` — mirroring each one's existing `pull_request` path filters. A merge to `main` carries the merged PR's changed files, so the relevant filtered workflow still fires on the commit that broke `main` (e.g. the runner incident: #349's `runner/**` files ⇒ runner test runs on the push to main). A broken `main` is now reported immediately after merge.
- The web test step runs the **full** suite on push to `main` (there is no PR base SHA for `--affected` on a `push` event) and stays scoped to affected packages on `pull_request`.
- Add `main` to `pull-request-build-lint-api.yml`'s PR branch list so ruff gates PRs into `main`, not just `preview`.

## Not in this PR (needs a human)

- **"Require branches to be up to date before merging"** on `main` — this is a branch-protection repo setting, not a repo file, so it can't ship in a PR. It's the direct fix for the semantic-merge-conflict class (at the cost of more rebases before landing). The `push:` triggers here report a broken `main` but don't *prevent* it; enabling that setting closes the prevention gap.
- **Auditing other path-filtered guards** for the same cross-cutting problem #362 hit (a check living in one tree that guards another) is worth a follow-up pass.

## Validation

- All four edited workflows parse as valid YAML.
- Trigger changes reviewed in isolation via `git diff`; no job logic changed except the web test step's conditional `--affected`.

### DIFF
--- a/.github/workflows/pull-request-build-lint-api.yml
+++ b/.github/workflows/pull-request-build-lint-api.yml
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
   pull_request:
     branches:
+      - "main"
       - "preview"
     types:
       - "opened"
--- a/.github/workflows/pull-request-test-api.yml
+++ b/.github/workflows/pull-request-test-api.yml
@@ -18,6 +18,16 @@ on:
       # runner/ and apps/web but not apps/api, so without this path the
       # guard would not run on the one PR shape it exists to catch.
       - "runner/src/config/schema.rs"
+  # Post-merge validation: PR checks don't re-run when the base branch moves,
+  # so a main built from two independently-green PRs may still be broken.
+  # Run pytest again on push to main to surface it immediately.
+  push:
+    branches:
+      - "main"
+    paths:
+      - "apps/api/**"
+      - ".github/workflows/pull-request-test-api.yml"
+      - "runner/src/config/schema.rs"
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
--- a/.github/workflows/pull-request-test-runner.yml
+++ b/.github/workflows/pull-request-test-runner.yml
@@ -15,6 +15,19 @@ on:
       - "Cargo.lock"
       - "runner/**"
       - ".github/workflows/pull-request-test-runner.yml"
+  # Post-merge validation: GitHub re-runs PR checks on synchronize, not when
+  # the base moves, so two PRs green against an older main can merge into a
+  # main that does not build. Run again on push to main to report that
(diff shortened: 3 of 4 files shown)


irichard00:
## Review: approved ✅

Config-only change; all five acceptance criteria verified against the head commit.

**Verified**
- `push: [main]` added to the three test workflows, path filters **mirror** each workflow's `pull_request` filters (runner: Cargo.toml/Cargo.lock/runner/**/workflow; api: apps/api/**/workflow/`runner/src/config/schema.rs`; web: the eight web paths + workflow).
- Web test step is conditional — `--affected` on `pull_request`, full `turbo run test` on push (correct: no PR base SHA on push).
- `pull-request-build-lint-api.yml` PR branches now `["main","preview"]`.
- All four files parse as valid YAML.

**Key correctness check — the jobs actually run on push.** The three test jobs have *no* job-level `if:` gate, so a `push: main` event runs them (unlike this repo's `lint-api` job, which is gated on `requested_reviewers != null`). If the test jobs had carried that gate, the trigger would have fired but every job would have been skipped and the fix would have been a no-op. Confirmed they don't. CI on this PR already shows `cargo test + clippy`, `pytest`, and `test` green.

**Non-blocking observations (out of this PR's scope)**
- `lint-api` still runs only once reviewers are requested (`if: ...requested_reviewers != null`). Adding `main` to the branch list matches the issue's ask and mirrors preview behavior, but ruff won't gate a *review-less* PR into main. Pre-existing, unchanged here — flagging for awareness.
- Prevention half (branch protection "Require branches to be up to date") and the broader path-filter/cross-tree guard audit remain open, as the author noted. This PR delivers the *reporting* half.

No changes requested.

PR #370, opened 2026-09-09
PDASHOSS01-139 State transitions: 15s debounce before dispatch, and supersede the in-flight run on a phase change

irichard00:
## What & why

Today a state change dispatches an agent run **synchronously and irrevocably**, and a second quick change is wasted or silently dropped. On `Todo → In Progress → In Review` a correcting drag can't stop the already-queued coding run, and the review run is swallowed by the single-active-run guardrail — leaving the impl run pushing commits on the wrong prompt and, on terminate, disarming the freshly review-armed ticker so the issue parks with no agent scheduled.

Fixes `PDASHOSS01-139`.

## Part 1 — 15s debounce

- New side table `IssuePendingDispatch` (one row per issue): monotonic `token`, `dispatch_at`, `from_state`/`to_state`.
- `signals.py` routes the `dispatch_immediate=True` path through the new `service.route_state_transition`. Entering a ticking state now records the intent and schedules `bgtasks.state_transition_debounce.dispatch_debounced_transition` with a 15s countdown instead of dispatching inline.
- A further transition inside the window bumps the token; the earlier delayed job no-ops. The job re-reads the issue under a row lock and dispatches against the **settled** state, so `Todo → In Progress → In Review` coalesces to one review-template run and `Todo → In Progress → Todo` yields no run and no armed ticker.
- `handle_issue_state_transition` stays the synchronous dispatcher (used by the debounce job, Comment & Run, project move, and tests). Issue **creation** into a ticking state is not debounced (`from_state is None`).

## Part 2 — phase-change supersede

- When a phase change lands while a run is in flight, `supersede_active_run` cancels it (immediately for QUEUED/PAUSED; via `CANCEL_REQUESTED` + a runner cancel frame otherwise), arms the new phase's ticker, and stashes a `_phase_change_handoff` marker — mirroring the cross-project move handoff.
- `complete_phase_change_handoff` creates the successor from the terminal callback once the old run stops, so the single-active-run guardrail / `agent_run_one_active_per_work_item` are never violated.
- The terminal-effects handoff guard is extended (`_has_phase_change_handoff`) so the cancelled run's `completed`/`blocked` done payload can't disarm the newly armed ticker. Reap / runner-revoke recovery paths call both completers (idempotent).

## Open question (deferred)

What to do when the new state is **non-ticking** (Paused/Done/Backlog/Cancelled) and a run is in flight. The issue reserved this for a human decision; implemented as **(a) leave the run running** (today's behavior, ticker disarms, no handoff) and flagged in an issue comment. Switching to (b)/(c) is a localized change in one branch of `route_state_transition`.

## Tests

- `tests/unit/orchestration/test_service.py`: debounce coalescing → one review run; cancel-on-leave → zero runs / no armed ticker; stale-token and issue-deleted no-ops; supersede (cancel-requested, marker, resume-parent capture); successor created on terminal with the ticker surviving the cancelled run's completed payload; QUEUED immediate supersede; non-ticking-exit leaves the run untouched.
- Updated `test_runs_views.py` (`test_no_skip_header_debounces_dispatch_on_state_change`) and `test_issue_move_run_repoint.py` to reflect the new signal wiring. Regression: the skip-immediate-dispatch header and Comment & Run tests still pass.
- Full `pi_dash/tests/unit/` suite: **1280 passed** (only the pre-existing environmental `test_valid_agents_matches_runner_schema`, which reads the repo-root `runner/` tree absent from the API test container, deselected). `ruff check` clean.

Design: `.ai_design/state_transition_debounce/design.md`.

### DIFF
--- a/.ai_design/state_transition_debounce/design.md
+++ b/.ai_design/state_transition_debounce/design.md
@@ -1,0 +1,187 @@
+# State-transition Debounce + Phase-change Supersede
+
+> Directory: `.ai_design/state_transition_debounce/`
+>
+> **Status:** implemented (PDASHOSS01-139).
+>
+> **Scope:** two changes to how an Issue state change drives an agent run.
+>
+> 1. **15s debounce** — a transition that targets a _ticking_ state no
+>    longer dispatches inline. It records the intent and fires after a 15s
+>    quiet period; a further transition inside the window replaces the
+>    intent and restarts the clock, so a rapid `Todo → In Progress → In
+Review` mis-drop coalesces into a single dispatch on the final state.
+> 2. **Phase-change supersede** — a phase change _after_ a run is already
+>    in flight cancels that run and hands off to the new phase's run once
+>    the old run terminates, mirroring the existing cross-project move
+>    handoff.
+
+Related designs: `.ai_design/issue_ticking_system/` (the per-issue
+ticker), `.ai_design/create_review_state/` (phase registry, fresh-session
+entry, terminal-signal disarm).
+
+## 1. Problem
+
+`orchestration/signals.py` fires `post_save(Issue)` →
+`orchestration/service.handle_issue_state_transition`, which armed the
+ticker and dispatched a run **synchronously, in the same request**. Two
+failures followed, both visible on `Todo → In Progress → In Review`:
+
+1. **No settling window.** Dragging a card to In Progress started a coding
+   run instantly. A correction to In Review seconds later could not stop
+   the already-queued coding run.
+2. **The second transition was swallowed.** On `In Progress → In Review`
+   the ticker was re-armed on the review cadence and `resume_parent_run`
+   was set, but the inline dispatch hit the single-active-run guardrail
+   (`service.py`) and the DB constraint `agent_run_one_active_per_work_item`
+   — so **no review run was created**. The impl run kept running on the
+   coding prompt while the card read In Review, and when it terminated its
+   done-payload disarmed the freshly review-armed ticker
+   (`maybe_disarm_on_terminal_signal`), parking the issue with no agent
+   scheduled at all.
+
+## 2. Design
+
+### 2.1 Where the debounce lives
+
+The debounce is layered in the **signal**, not in
+`handle_issue_state_transition`. That function stays the synchronous
+dispatcher used directly by tests, by the debounce job, by Comment & Run
+(`dispatch_immediate=False`), and by the project-move handoff. The signal's
+`dispatch_immediate=True` path now routes through
+`service.route_state_transition`.
+
+`route_state_transition` decides:
+
+| condition                                                    | action                                                                                                       |
+| ------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------------ |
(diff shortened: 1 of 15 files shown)


irichard00:
## Review — PDASHOSS01-139 (debounce + phase-change supersede)

Reviewed the full diff against the issue's acceptance criteria. **The implementation is sound and I recommend approval.** It faithfully mirrors the existing cross-project-move handoff, the concurrency handling is careful, and the test coverage maps to every acceptance criterion.

### Verified against acceptance criteria
- **`Todo → In Progress → In Review` within 15s → one review-template run.** Traced: the first move debounces (token 1, no inline arm); the second move (no active run yet, so not a supersede) re-schedules (token 2, `from_state=In Progress`); the token-1 job no-ops on stale token; the token-2 job fires against the settled state and `handle_issue_state_transition(In Progress→In Review)` renders the review template with `fresh_session`. ✔ (`test_debounce_coalesces_rapid_transitions_into_one_review_run`)
- **`Todo → In Progress → Todo` within 15s → zero runs, no armed ticker.** The non-ticking target clears the pending token and disarms; no ticker was ever armed during the window. ✔
- **Supersede after dispatch → impl run `cancel_requested`, review run created on terminal, ticker survives the cancelled run's `completed` payload.** The handoff guard in `apply_terminal_effects` now covers the phase-change kind, so `_apply_post_run_orchestration` (the terminal-signal disarm) is suppressed. ✔ (`test_phase_change_successor_created_on_terminal_and_ticker_survives` — non-vacuous; asserts CANCEL_REQUESTED, fresh successor, `ticker.enabled`, and marker idempotency)
- **Comment & Run / skip-immediate-dispatch unchanged.** `dispatch_immediate=False` never enters the router. ✔ (regression tests updated, not weakened)
- **Single-active-run + `agent_one_active_per_work_item` never violated.** The successor is only created from the terminal callback (or in the same txn for an inert QUEUED/PAUSED run). ✔
- **Deletion / cross-project move inside the window cancels dispatch.** `issue-gone` / `stale-token` no-ops. ✔

### One non-blocking observation (durability)
The debounce **defers ticker arming** until the countdown job fires — during the 15s window the issue sits in a ticking state with *neither* an armed ticker *nor* a pending run. If the Celery countdown message is lost (broker restart, or a worker outage spanning the window), nothing recovers it: there is no periodic sweep over `IssuePendingDispatch.dispatch_at`, and the ticker that would otherwise self-heal on the next tick was never armed. Pre-change, the ticker armed synchronously so a dropped immediate dispatch self-healed.

Probability is low and this is an inherent tradeoff of debounce-before-arm, so it needn't block the PR — but it's a genuine regression in the failure mode and isn't called out in the design doc. Worth a follow-up: either a periodic reconciliation of overdue `dispatch_at` rows, or arm the ticker at schedule time.

### Open question (needs a human call — already surfaced)
The non-ticking-exit-with-run-in-flight question is shipped as option (a) (run continues, ticker disarms) and flagged for a product decision in the issue thread. Not a review defect — just noting the default is deliberate and reversible in one branch of `route_state_transition`.

PR #376, opened 2026-09-11
PDASHOSS01-155 pidash CLI: auto-install missing agent via its official install script when adding a runner

irichard00:
## What & why

`PDASHOSS01-155`. When `pidash runner add` registers a runner whose selected
agent CLI isn't installed on the machine, Pi Dash now **installs it by invoking
the vendor's own official install script** instead of only opening the install
page (the prior PDASHOSS01-13/-15 behaviour). The runner then reaches a ready
state after the one login step installers can't do.

**Key principle preserved:** Pi Dash never bundles, mirrors, pins, or ships an
agent binary. We only call the vendor's own script at runtime, so hosts always
track the vendor's latest supported build and there's no redistribution/licensing
concern.

## How it works

On `runner add`, after enrollment + service setup succeed, for the selected agent:

1. Probe binary presence via the same `util::shell::binary_runs_version` helper
   the doctor path uses. If present, do nothing.
2. If missing, print which agent is missing and the **exact official command**
   about to run, then run it with output streamed to the terminal
   (`curl … | bash` on macOS/Linux/WSL, PowerShell on Windows).
3. Re-detect the binary and resolve its **absolute path** — installers usually
   drop it in a dir like `~/.local/bin` not yet on this process's PATH — then
   write that path into the runner's agent-section `binary` config.
4. Print the one login step still required (installers don't authenticate).

Never hard-fails: on any install error (no network, no scriptable installer for
that agent), it prints the error, falls back to opening the vendor install page,
and leaves `pidash doctor` reporting the agent as missing.

## Agent → official install command (single data table)

Lives in one place: `AgentKind::official_installer()` in `schema.rs`. Each row
links the vendor doc it was verified against. Only official vendor domains.

| Agent | macOS/Linux/WSL | Windows | Verified against |
| --- | --- | --- | --- |
| `claude-code` | `curl -fsSL https://claude.ai/install.sh \| bash` | `irm https://claude.ai/install.ps1 \| iex` | code.claude.com/docs/en/setup |
| `codex` | `curl -fsSL https://chatgpt.com/codex/install.sh \| sh` | `npm install -g @openai/codex` | learn.chatgpt.com/docs/codex/cli |
| `cursor-agent` | `curl https://cursor.com/install -fsS \| bash` | `irm 'https://cursor.com/install?win32=true' \| iex` | cursor.com/docs/cli/installation |
| `open-claw`, `grok`, `muse-code` | — (open-install-page fallback) | — | no verified vendor script yet |

## Opt-out & non-interactive

- `--skip-agent-install` skips the install step for operators who manage installs
  themselves.
- The vendor scripts are non-interactive, so CI shells don't hang; the
  browser-open fallback only fires when attached to a terminal.

## Changes

- `runner/src/config/schema.rs` — `AgentInstaller` struct + `official_installer()`
  data table + `post_install_login_hint()`; unit tests (mapping, platform
  selection, https/official-domain enforcement, login hints).
- `runner/src/cli/agent_install.rs` (new) — install execution, absolute-path
  resolution, config persistence; unit tests.
- `runner/src/cli/runner.rs` — `ensure_agent_installed` flow replacing the
  open-page-only reminder; `--skip-agent-install` flag.
- `runner/src/cli/mod.rs` — register module.
- Docs: `runner/README.md` (new "Agent CLI auto-install" section) + CLI reference.

## Tests

- `cargo build`, `cargo clippy --all-targets` — clean.
- `cargo test --lib` — 467 pass; `cargo test --test pidash_runner_add_auth` — 4 pass.

### DIFF
--- a/docs/wiki/17-cli-reference.md
+++ b/docs/wiki/17-cli-reference.md
@@ -82,16 +82,21 @@ pidash runner add --project <PROJECT> [--name <NAME>] [--workspace <SLUG>]
                   [--pod <POD>] [--working-dir <PATH>] [--agent codex|claude-code]
 ```
 
-| Flag                | Purpose                                                            |
-| ------------------- | ------------------------------------------------------------------ |
-| `--project <P>`     | Project identifier (slug or UUID). **Required.**                   |
-| `--name <N>`        | Human-friendly runner name. Auto-generated if omitted.             |
-| `--workspace <S>`   | Workspace slug. Required if you belong to multiple workspaces.     |
-| `--pod <P>`         | Pod within the project. Defaults to project's default pod.         |
-| `--working-dir <P>` | Local working dir for clones. Defaults to a path under `data_dir`. |
-| `--agent <K>`       | `codex` (default) or `claude-code`.                                |
+| Flag                     | Purpose                                                                                                                |
+| ------------------------ | ---------------------------------------------------------------------------------------------------------------------- |
+| `--project <P>`          | Project identifier (slug or UUID). **Required.**                                                                       |
+| `--name <N>`             | Human-friendly runner name. Auto-generated if omitted.                                                                 |
+| `--workspace <S>`        | Workspace slug. Required if you belong to multiple workspaces.                                                         |
+| `--pod <P>`              | Pod within the project. Defaults to project's default pod.                                                             |
+| `--working-dir <P>`      | Local working dir for clones. Defaults to a path under `data_dir`.                                                     |
+| `--agent <K>`            | Which agent CLI the runner drives: `codex` (default), `claude-code`, `cursor-agent`, `open-claw`, `grok`, `muse-code`. |
+| `--model <M>`            | Default LLM model for the agent. Omit to use the agent's own default.                                                  |
+| `--reasoning-effort <T>` | Codex reasoning tier (`low`/`medium`/`high`/`xhigh`). `--agent codex` only.                                            |
+| `--skip-agent-install`   | Don't auto-install the agent CLI if it's missing; you'll install it yourself.                                          |
 
 On the first runner: installs the OS service (systemd user unit / launchd agent / Windows scheduled task) and starts the daemon.
+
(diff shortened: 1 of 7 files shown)


irichard00 (commented):
## Code review — PDASHOSS01-155 (auto-install missing agent)

Reviewed the full diff (`schema.rs` data table, `agent_install.rs`, `runner.rs` wiring, tests, docs). Verified locally on this branch: `cargo clippy --all-targets` clean, the new unit tests pass (`agent_install` 4/4; schema installer/platform/https-domain/login-hint tests pass).

**Looks good:**
- Single data table in `AgentKind::official_installer()` — one `AgentInstaller` row per agent, vendor doc URL linked per row, official-domain + https enforced by a unit test. Adding an agent is one row.
- Never bundles a binary; only invokes the vendor script at runtime.
- Presence check reuses the same `login_shell_command` path as `doctor.rs::check_version` (via `binary_runs_version`), so "already installed → skip" and doctor's report stay consistent.
- Absolute-path resolution after install (login-shell `command -v` first, then the `~/.local/bin` candidate fallback) correctly avoids trusting the current process PATH, and persists the resolved path into the matching agent section.
- Non-fatal by construction: every branch falls back to the install page and returns `()`; `runner add`'s result is discarded at the call site, so a stale in-memory `RunnerConfig.binary` can't mislead anything.
- `--skip-agent-install` opt-out, post-install login hints, and README/wiki docs all present.

**Non-blocking observations:**
1. **Non-interactive/CI auto-install is ungated.** `ensure_agent_installed` executes `curl … | bash` whenever the binary is missing and `--skip-agent-install` isn't set, with no interactivity check (only `open_install_page` gates on `is_terminal`). The vendor scripts are non-interactive so this won't *hang* (the literal acceptance criterion is met), but silently downloading+executing a remote install script in a non-interactive/CI shell is a surprising default. Consider skipping auto-install (and just printing the command/page) when neither stdin nor stdout is a TTY, leaving `--skip-agent-install` for the explicit case. Judgment call — flagging for a maintainer decision, not requesting a change.
2. **`bash -lc` vs the installer's PATH line.** `which_via_login_shell` uses a login shell (`-l`), while vendor installers typically append PATH to `.bashrc` (interactive-only). For the three supported agents this is harmless because they all drop into `~/.local/bin`, which the candidate-path fallback probes; worth a one-line comment noting the resolution leans on that fallback rather than the appended PATH.

**Not verifiable in this session:** the live end-to-end install on a genuinely clean macOS/Linux/Windows host and the real Windows PowerShell path — same constraint the implementer documented. Those are the OPS steps for the In Test phase.

No correctness defects found. LGTM pending the OPS checks.

PR #378, opened 2026-09-12
PDASHOSS01-118 [MR-10] Whitelist engine_version → dev_metadata.codex_version in session-open

irichard00:
## PDASHOSS01-118 — [MR-10] Whitelist engine_version → dev_metadata.codex_version in session-open

Design refs: managed_runner §12.3 (last bullet), §15 (Support row), §15.1.

The `engine_version` → `dev_metadata.codex_version` whitelist (≤64 chars) in `session_service._merge_dev_metadata` already landed alongside the `managed_runner` executor. This PR completes the remaining MR-10 items:

- **Log event** — `apply_hello` now emits a `managed_runner.engine_version` structured event (runner id + capped version) at session open, following the `managed_runner.*` logger conventions. It fires only when the runner reports a non-empty `engine_version` this open, so legacy / non-managed daemons that never send it produce no noise.
- **Admin runner list** — `RunnerAdmin` surfaces support's "which build is this user on" signals: adds `provisioning` and `runner_version` columns, a `codex_version` column reading `dev_metadata.codex_version`, and a `provisioning` list filter.

The runner side sends the value in MR-12 (out of scope here).

## Testing

- `pi_dash/tests/unit/managed_runner/test_serializers_and_model.py` — 18 passed (includes 2 new: log emitted when reported, omitted when not).
- `pi_dash/tests/unit/runner/test_runner_dev_metadata.py` — 3 passed.
- `python manage.py check` — no issues (validates the admin `list_display` method column).

### DIFF
--- a/apps/api/pi_dash/runner/admin.py
+++ b/apps/api/pi_dash/runner/admin.py
@@ -44,13 +44,25 @@ class RunnerAdmin(admin.ModelAdmin):
         "dev_machine",
         "visibility",
         "status",
+        "provisioning",
+        "runner_version",
+        "codex_version",
         "host_label",
         "refresh_token_generation",
         "last_heartbeat_at",
         "revoked_at",
     )
-    list_filter = ("status", "visibility", "workspace")
+    list_filter = ("status", "visibility", "provisioning", "workspace")
     search_fields = ("name", "owner__email", "host_label")
+
+    @admin.display(description="codex version")
+    def codex_version(self, obj):
+        # Support's "which build is this user on" signal (design §15.1). The
+        # engine version is whitelisted into ``dev_metadata.codex_version`` at
+        # session open (see ``session_service._merge_dev_metadata``).
+        if isinstance(obj.dev_metadata, dict):
+            return obj.dev_metadata.get("codex_version") or "—"
+        return "—"
     readonly_fields = (
         "refresh_token_hash",
         "refresh_token_fingerprint",
--- a/apps/api/pi_dash/runner/services/session_service.py
+++ b/apps/api/pi_dash/runner/services/session_service.py
@@ -126,6 +126,18 @@ def apply_hello(runner: Runner, body: Dict[str, Any]) -> None:
     runner.runner_version = body.get("version", "") or runner.runner_version
     runner.dev_metadata = _merge_dev_metadata(runner.dev_metadata, body)
     runner.last_heartbeat_at = timezone.now()
+    # Observability (design §15.1): a bundled runner ships its agent binary
+    # inside the desktop app, so the engine version it self-reports at session
+    # open is support's only signal for "which build is this user on". Emit it
+    # only when the runner actually reports a non-empty value this open — a
+    # legacy or non-managed daemon that never sends it produces no noise.
+    reported_engine_version = body.get("engine_version")
+    if isinstance(reported_engine_version, str) and reported_engine_version:
+        logger.info(
+            "managed_runner.engine_version runner=%s version=%s",
+            runner.id,
+            reported_engine_version[:64],
+        )
     update_fields = [
         "os",
         "arch",
--- a/apps/api/pi_dash/tests/unit/managed_runner/test_serializers_and_model.py
+++ b/apps/api/pi_dash/tests/unit/managed_runner/test_serializers_and_model.py
@@ -209,6 +209,34 @@ def test_engine_version_is_length_capped(bundled_runner):
     assert len(bundled_runner.dev_metadata["codex_version"]) <= 64
 
 
+def test_apply_hello_emits_engine_version_log_event(bundled_runner, caplog):
+    """Design §15.1: a ``managed_runner.engine_version`` structured event is
+    logged at session open so support can see which bundled build reported in."""
+    import logging
+
+    from pi_dash.runner.services.session_service import apply_hello
+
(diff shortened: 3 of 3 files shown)


irichard00:
**Review: approved ✅**

Reviewed the MR-10 changes against the acceptance criteria and design §15.1. All four parts land correctly.

- **Whitelist** (`_merge_dev_metadata`, already on main): caps `engine_version` at 64 chars, clears `codex_version` on empty string, ignores non-string / unknown keys. Confirmed.
- **Log event** (`apply_hello`): emits `managed_runner.engine_version` at INFO with `runner=<id>` and the `[:64]`-capped version, guarded by `isinstance(str) and non-empty` — so legacy/non-managed daemons stay silent. Guard mirrors the persistence path. ✓
- **Admin columns**: `provisioning`, `runner_version`, and a `codex_version` method column reading `dev_metadata.codex_version` with a `—` fallback (and a dict guard). ✓
- **Admin filter**: `provisioning` added to `list_filter`. ✓

**Verification (re-ran locally, not just trusting the handoff):**
- `pytest test_serializers_and_model.py test_runner_dev_metadata.py` → **21 passed** (includes the 2 new emit/omit log tests).
- `python manage.py check` → no issues (confirms the `codex_version` method column resolves).

No blocking issues. Runner-side send of `engine_version` arrives in MR-12, as noted.

irichard00:
**Review: 1 finding, fixed in-place ✅**

Reviewed at `6ba17b00`. The previous review pass on this PR approved an earlier commit and was false-green (it never checked that the log event reached a handler), so I re-reviewed the whole diff from scratch rather than only the delta.

### The three production changes hold up

- **`session_service.py`** — `event_logger = getLogger("pi_dash.managed_runner")` with the reasoning in a comment. Emission is guarded on `isinstance(v, str) and v`, capped at `[:64]` to match what `_merge_dev_metadata` persists, and `%`-style args are correct. `apply_hello` has exactly one caller (`runner/views/sessions.py:191`, the session-open path), so the event fires at session open and nowhere else — no heartbeat/poll noise.
- **`settings/local.py` / `production.py`** — `pi_dash.managed_runner` declared with the `console` handler; prod uses `DEBUG if DEBUG else INFO` and the `console` handler is `INFO`, so the record gets through. This is the right fix for the root cause: `disable_existing_loggers: True` + no `root` logger means an undeclared name gets zero handlers.
- **`admin.py`** — `provisioning` and `runner_version` are real `Runner` fields, `codex_version` is a valid `@admin.display` method, and the module comment correctly states that none of it is reachable (no `django.contrib.admin`, no `admin/` route). That half is split to PDASHOSS01-209; nothing here to fix.

### ❌ Finding — the new regression test leaks logging config into the rest of the suite

`test_engine_version_event_reaches_a_handler_under_real_logging_config` calls `logging.config.dictConfig()` with the project's own `LOGGING`. `dictConfig` rewires **every** logger the config declares, not just `pi_dash.managed_runner` — and the `finally` block restored only that one. So `pi_dash.api`, `pi_dash.api.request`, `pi_dash.worker`, `pi_dash.exception` and the rest keep the test's in-memory `_Capture` handler and `propagate=False` for the remainder of the pytest process, and keep appending records to a list owned by a test that already finished.

Observed directly:

```
--- BEFORE
  pi_dash.api                  handlers=[] propagate=True
  ...
--- AFTER (only pi_dash.managed_runner was restored)
  pi_dash.api                  handlers=['_Capture'] propagate=False
  pi_dash.api.request          handlers=['_Capture'] propagate=False
  pi_dash.worker               handlers=['_Capture'] propagate=False
  pi_dash.exception            handlers=['_Capture'] propagate=False
  pi_dash.managed_runner       handlers=[] propagate=True

leaked into the test's closure list after the test finished: 1 record(s)
```

Concrete failure: a one-line test asserting on `caplog` for `pi_dash.api` **passes alone** and **fails when run after this file**, because pytest's caplog handler lives on the root logger and `propagate=False` stops records from ever getting there.

```
### probe ALONE:                          1 passed
### probe AFTER the PR's test file:       1 failed, 21 passed
```

This is the same class of trap the PR is fixing — a logging test that looks green but is lying about the process state.

**Fixed in `9009edc5`**: snapshot `level` / `handlers` / `propagate` / `disabled` for every logger the config declares and restore them all in the `finally`. Same probe after the fix: `22 passed`.

### Gates re-run (not taken from the hand-off)

| Check | Result |
|---|---|
| `pytest test_serializers_and_model.py test_runner_dev_metadata.py` | **24 passed** |
| `pytest unit/managed_runner/ unit/runner/` | **532 passed, 1 failed** |
| `ruff check` on all 5 changed files | clean |

The one failure is `test_machine_commands.py::test_valid_agents_matches_runner_schema` — `IndexError: 6` from `pathlib.Path(__file__).resolve().parents[6] / "runner/src/config/schema.rs"`, which cannot resolve because only `apps/api` is mounted, not the repo root. Pre-existing environment artifact; this PR touches no path logic.

### One thing to know before merging

`pi-dash/pdashoss01-119` (MR-11) independently adds the **same** `pi_dash.managed_runner` logger block to both settings files, and its `test_observability.py` copies this test's `dictConfig` pattern — including the same restore-only-one-logger bug. Whichever branch merges second will conflict in `local.py`/`production.py`, and MR-11's copy of the test still needs the same fix.

No other findings. Approving.

PR #379, opened 2026-09-12
PDASHOSS01-119 [MR-11] managed_runner/checks.py startup checks and observability events/metrics

irichard00:
## What

MR-11 of the Managed Runner epic (PDASHOSS01-107). Fills the remaining observability gaps for the `managed_runner` executor kind, per design `.ai_design/managed_runner/design.md` §15.1 / §17.

The startup checks (`managed.E001` / `E002`) and most log events (`enrolled`, `removed`, `token_issued` / `token_refresh_failed`, `engine_version`, `queued_expired`) already landed in earlier rows. This PR adds the two remaining log events and the three metrics.

## Log events

A single `post_save(AgentRun)` handler in `managed_runner/signals.py` (wired from `runner/apps.py:ready()` next to the existing checks import) emits, for a **newly created** managed run — the one place the run id and resolved runner exist together, covering all five AgentRun creation sites uniformly:

- `managed_runner.queued_waiting run=<id> runner=<id>` — an automatic run parked for a closed desktop (`error_code=desktop_not_connected`, §8.5)
- `managed_runner.run_pinned run=<id> runner=<id>` — a run pinned to a live desktop

Follows the house convention (dotted event name + `key=%s`, `logging.getLogger(__name__)`). Never logs a token or a home-directory path.

## Metrics (`GET /api/v1/runner/metrics/`)

Added in the endpoint's existing hand-rolled Prometheus text idiom — no new dependency, each series a DB snapshot at scrape time:

- `pi_dash_managed_runner_online` (gauge) — online desktop-bundled runners
- `pi_dash_runs_total{executor_kind="…"}` (counter) — AgentRuns by executor kind, zero-filled for every known kind
- `pi_dash_managed_runner_queued_wait_seconds` (histogram) — age of managed runs currently `QUEUED` waiting for a desktop

## Tests

- `tests/unit/runner/test_metrics.py` — endpoint output for each new series (establishes the metrics-endpoint test pattern, which did not exist before)
- `tests/unit/managed_runner/test_observability.py` — both events fire correctly; non-managed runs and plain updates emit nothing

`manage.py check` passes (E001/E002 still register). Full `managed_runner` suite (103) and `runner`/`orchestration` suites green; the one unrelated pre-existing failure (`test_valid_agents_matches_runner_schema`) only reproduces in the apps/api-only test container because it reads a Rust file outside the mount.

### DIFF
--- a/apps/api/pi_dash/managed_runner/signals.py
+++ b/apps/api/pi_dash/managed_runner/signals.py
@@ -1,0 +1,56 @@
+# Copyright (c) 2023-present Pi Dash Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
+"""Structured observability events for managed-runner run creation.
+
+Two of the events in design §15.1 — ``managed_runner.run_pinned`` and
+``managed_runner.queued_waiting`` — describe a run the moment it is created, but
+the pinning decision is made in ``cloud_agent/creation.py`` *before* the row
+exists, and the fields it returns are splatted into ``AgentRun.objects.create``
+by five separate creation sites. A single ``post_save`` handler is the one place
+that sees both the run id and the resolved runner together, for every path,
+without threading a log call through each caller.
+
+Every managed run that is admitted carries a ``pinned_runner`` (an online one
+when available, the enrolled-but-offline one when an automatic run is parked to
+wait). The two cases are told apart by ``error_code``: a parked run carries
+``desktop_not_connected`` (``ManagedRunnerReason.NOT_CONNECTED``); a run pinned
+to a live desktop carries none.
+"""
+
+from __future__ import annotations
+
+import logging
+
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from pi_dash.core.agent_execution import AgentExecutorKind
+from pi_dash.managed_runner.errors import ManagedRunnerReason
+from pi_dash.runner.models import AgentRun
+
+logger = logging.getLogger(__name__)
+
+
+@receiver(post_save, sender=AgentRun)
+def log_managed_run_pinning(sender, instance: AgentRun, created: bool, **kwargs):
+    """Emit ``run_pinned`` / ``queued_waiting`` for a newly created managed run.
+
+    A no-op for every other executor kind and for updates, so the cost on the
+    hot run-creation path is a single attribute comparison.
+    """
+    if not created or instance.executor_kind != AgentExecutorKind.MANAGED_RUNNER:
+        return
+    if instance.error_code == ManagedRunnerReason.NOT_CONNECTED:
+        logger.info(
+            "managed_runner.queued_waiting run=%s runner=%s",
+            instance.id,
+            instance.pinned_runner_id,
+        )
+    elif instance.pinned_runner_id is not None:
+        logger.info(
+            "managed_runner.run_pinned run=%s runner=%s",
+            instance.id,
+            instance.pinned_runner_id,
+        )
--- a/apps/api/pi_dash/runner/apps.py
+++ b/apps/api/pi_dash/runner/apps.py
@@ -17,3 +17,7 @@ class RunnerConfig(AppConfig):
         from pi_dash.runner import signals  # noqa: F401
         from pi_dash.cloud_agent import checks  # noqa: F401
         from pi_dash.managed_runner import checks as managed_checks  # noqa: F401
+
+        # Registers post_save(AgentRun) handler emitting managed_runner
+        # run_pinned / queued_waiting observability events (design §15.1).
+        from pi_dash.managed_runner import signals as managed_signals  # noqa: F401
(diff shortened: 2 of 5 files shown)


irichard00 (commented):
### Review: MR-11 managed_runner observability (checks / events / metrics)

Reviewed the diff against design §15.1/§17 and the issue's acceptance criteria. Verified against the codebase (not just the diff):

**Correct — no blocking issues found.**

- `managed_runner/signals.py` — the single `post_save(AgentRun)` handler is the right call. Confirmed that `_managed_execution_fields` (`cloud_agent/creation.py`) sets both `pinned_runner` **and** `error_code=desktop_not_connected` into the `fields` dict that every creation site splats via `AgentRun.objects.create(**execution)`, so `post_save(created=True)` sees both — `run_pinned` vs `queued_waiting` are correctly distinguished at creation time. `ManagedRunnerReason.NOT_CONNECTED == "desktop_not_connected"` (plain str) compares correctly against the `error_code` CharField. Neither event logs a token or a home path (only `run.id` / `pinned_runner_id`). ✅
- `runner/views/metrics.py` — `pi_dash_managed_runner_online` (DESKTOP_BUNDLED + ONLINE), `pi_dash_runs_total{executor_kind}` (zero-filled over `AgentExecutorKind.values`), and the `pi_dash_managed_runner_queued_wait_seconds` histogram (cumulative buckets, `+Inf` == `_count`, managed+QUEUED only) all match the AC. Pre-existing gauges are preserved. Prometheus text format is well-formed (HELP/TYPE per family, ascending `le`, trailing newlines). ✅
- Tests — endpoint text assertions and the signal `caplog` assertions cover the AC directly, including the negative cases (manual/local runners excluded, updates not logged, non-managed/non-queued excluded from the histogram). Fixtures (`bundled_runner`, `project`, `create_user`, `api_client`) exist. ✅

**Non-blocking note (design tradeoff, already documented in-code):**
`pi_dash_runs_total` is exposed as `# TYPE … counter` but is a scrape-time `COUNT(*)` from the DB. `AgentRun.workspace` is `on_delete=CASCADE`, so deleting a workspace drops rows and the count can decrease — Prometheus would read that as a counter reset and mis-attribute `rate()`. The code comment acknowledges "DB rows are never deleted in normal operation," and the "snapshot / no new deps" approach is a deliberate, reasonable choice for this endpoint, so this is a caveat to be aware of rather than a change request. If accurate `rate()` over the managed-runner rollout ever matters, a persisted counter would be needed.

Approving; leaving the state to the human. Validation (per handoff): 9 targeted tests + 103 managed_runner + 563 runner/orchestration pass; `manage.py check` clean.

irichard00:
Pushed `d6676db8` to close the one defect the test pass found (AC 2 and 3).

`run_pinned` and `queued_waiting` were computed correctly but never written
anywhere in a running instance: `LOGGING` sets `disable_existing_loggers: True`
and declares no `root` logger, so `getLogger("pi_dash.managed_runner.signals")`
resolved to WARNING with zero handlers and the INFO record was discarded before
formatting. The `caplog` tests were false-green — `caplog.at_level` attaches its
own handler and forces the level, so it is structurally blind to this.

What changed:

- New `managed_runner/events.py` holds the single `event_logger`
  (`pi_dash.managed_runner`) with the rationale. Every `managed_runner.*` event
  now goes through it: `run_pinned`/`queued_waiting` (signals), `queued_expired`
  (tasks), `enrolled`/`removed` (runner desktop view), `token_issued`/
  `token_refresh_failed` (assistant agent-profile view). Each of those modules
  used `getLogger(__name__)` and every logging call in them is one of these
  events, so no unrelated module's INFO output is turned on.
- `pi_dash.managed_runner` declared at INFO on the existing `console` handler in
  both `settings/production.py` and `settings/local.py`. Scoped to the event
  namespace rather than a `pi_dash` parent — the repo-wide version of this gap
  stays with PDASHOSS01-208.
- Tests: the declaration is asserted against the real `LOGGING` dicts; a
  delivery test applies the project's own config (handlers swapped for an
  in-memory one) and asserts the `run_pinned` record actually arrives; one more
  asserts no event payload carries a token or a home-directory path.

Note on the settings hunk: it is byte-identical to the one #378 (MR-10) adds for
`engine_version`, deliberately, so the two merge cleanly —
`git merge-tree --write-tree` on the two branches succeeds with no conflict.

Verified in the API container under `pi_dash.settings.production`: the event
logger reports `isEnabledFor(INFO)` with a `StreamHandler`, and driving the
`post_save` handler with the two managed row shapes emits

```
{"levelname":"INFO","name":"pi_dash.managed_runner","message":"managed_runner.run_pinned run=… runner=…"}
{"levelname":"INFO","name":"pi_dash.managed_runner","message":"managed_runner.queued_waiting run=… runner=…"}
```

while a `local_runner` row logs nothing. `manage.py check` clean (and still
reports `managed.E001` when `DEFAULT_AGENT_EXECUTOR=managed_runner` meets
`MANAGED_RUNNER_ENABLED=false`). `managed_runner` + `runner/test_metrics.py` =
112 passed; `runner` + `orchestration` = 563 passed with the one pre-existing
environmental failure (`test_machine_commands` reads a Rust file that is not
mounted in the apps/api-only test container).

PR #384, opened 2026-09-12
PDASHOSS01-161 Two-tier .pidash/context.md: machine-level workspace context + per-working-dir project context

irichard00:
## Summary

Splits `.pidash/context.md` into two tiers (PDASHOSS01-161):

- **Machine-level** `~/.pidash/context.md` (`scope: "machine"`) — every project in every bound workspace, keyed by workspace slug. Written best-effort on `auth login` and rewritable with the new `pidash context refresh`. Location follows `PIDASH_CONFIG_DIR` / `--config-dir`, else `~/.pidash`, else the config dir (**D1**), so isolated installs (desktop, CI) never write into the real home directory.
- **Per-working-dir** `<working_dir>/.pidash/context.md` (`scope: "project"`) — the single project this dir belongs to. Now **replaces** rather than upserts (fixing the one-way accumulation) and carries `repo_url`.

Both files carry `scope` + a `generated_at` RFC 3339 UTC stamp; each project carries `repo_url` (already returned by the projects API). The hand-rolled parser is refactored into **one indentation-aware line scanner** shared by both documents (**D4**), tolerating legacy files with no `scope` / `generated_at` / `repo_url`. The working-dir file is excluded from git via `$GIT_DIR/info/exclude` — handling both a `.git` directory and a `.git`-file worktree pointer — never `.gitignore` (**D5**). Machine file is keyed by workspace and upserted by slug (**D3**).

Prompt fragments (`fragments/03_pidash_cli.md` + `sections/pidash-cli.md`) and the CLI reference now describe both files and the precedence between them.

## Scope note

The `pi-dash-skill/` and `private-pi-dash/` skill trees named in the issue are **not present in this OSS repo** — only the API prompt fragments and CLI docs are here. Those were updated; the private plugin copies must be updated in their own repo.

## Testing

- `cargo test --locked` (runner) — full suite green; new tests: D1 path override/default branches, project/machine round-trips, legacy-file parse, workspace upsert-by-slug, working-dir replace, git-exclude (dir + `.git`-file + non-repo).
- `cargo clippy --all-targets --locked -- -D warnings` — clean.
- API prompting unit tests (`test_registry`, `test_validation`, `test_composer`, `test_recipes`) — 72 passed.

### DIFF
--- a/apps/api/pi_dash/prompting/fragments/03_pidash_cli.md
+++ b/apps/api/pi_dash/prompting/fragments/03_pidash_cli.md
@@ -11,6 +11,15 @@ The CLI reads the following from the process environment — never pass them as 
 - `PIDASH_TOKEN` — session-scoped credential. Treat it like any other secret.
 - `PIDASH_ISSUE_IDENTIFIER` — the current issue identifier (`{{ issue.identifier }}`). When set, `pidash state list` defaults to this issue's project so you can call it with no args.
 
+### Working context files (`context.md`)
+
+Two `context.md` files on the dev machine describe the Pi Dash projects available to you. Both carry a `generated_at` RFC 3339 UTC stamp so you can judge staleness, and each project entry carries a `repo_url` you can match against a local checkout.
+
+- **Working-dir** `<working_dir>/.pidash/context.md` (`scope: "project"`) — the single project this working directory belongs to. This is **authoritative for "which project am I working in"**.
+- **Machine-level** `~/.pidash/context.md` (`scope: "machine"`) — every project in every workspace this machine is bound to, grouped under a `workspaces:` list keyed by slug. This is the **lookup table for "what else exists in this workspace"** — use it to find a related issue's project or file a follow-up in a different project — and the **fallback** when there is no working-dir file.
+
+Precedence: trust the working-dir file for your current project; consult the machine-level file for everything else. If the machine-level file looks stale or is missing, run `pidash context refresh` to rewrite it for the bound workspace. (Its path follows `PIDASH_CONFIG_DIR` when set, otherwise `~/.pidash`.)
+
 ### Output contract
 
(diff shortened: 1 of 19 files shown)


irichard00:
### Review — PDASHOSS01-161 two-tier `context.md`

Reviewed the diff against the issue and its decisions D1–D5. Overall this is solid: the parser is single-sourced (D4) with an indentation-aware scanner, the working-dir write replaces rather than accumulates (D3), the D1 path logic is correct with tests for both branches, and coverage is good (round-trip for both shapes, legacy file, upsert-by-slug, git-exclude for `.git` dir / `.git`-file pointer / non-repo). All acceptance checkboxes on the issue are satisfied.

**One divergence from D2 worth a decision (non-blocking):**

D2 says *"The machine file is rewritten on auth login, on runner add, and on daemon workspace resolution, and a new `pidash context refresh` subcommand…"*. In this PR the **machine-level** file is written only on `auth login` (`cli/auth/login.rs:160`) and `pidash context refresh`. Both `runner add` (`cli/runner.rs:276`) and daemon workspace resolution (`daemon/supervisor.rs:2709`) call only `write_context_for_project` (the working-dir writer), not `write_machine_context`.

`docs/wiki/17-cli-reference.md` documents the narrower behavior consistently, so this is a deliberate narrowing rather than an oversight. Practical impact is limited — login populates the machine file per workspace and `context refresh` handles staleness — but a runner added to a workspace after login (or projects added/removed in a bound workspace) won't be reflected in `~/.pidash/context.md` until a re-login or manual `refresh`, which is precisely the multi-workspace freshness case D3 exists for.

Please confirm the intended trigger set: either wire `write_machine_context` into the `runner add` / daemon paths to match D2, or record that login + explicit `refresh` is the accepted scope and treat D2's extra triggers as descoped.

No correctness issues found in the parser/renderer or git-exclude logic.

PR #386, opened 2026-09-13
PDASHOSS01-156 Drop dead worktree-pool columns: Runner.free_worktrees and AgentRun.queue_position

irichard00:
## Summary

The worktree pool is retired (PDASHOSS01-134 removed it from the daemon, PDASHOSS01-137 retired the cloud-side half). Two columns added by `0017_worktree_pooling` outlived it as dead weight with no live readers or writers. This drops both and everything that existed solely to carry them.

- **`Runner.free_worktrees`** — per-runner capacity hint. No longer written or read; the matcher ranks eligible idle runners by `-last_heartbeat_at` alone.
- **`AgentRun.queue_position`** — display-only local-queue position. Every remaining write only cleared it to `None`.

## Changes

- **Model** (`runner/models.py`): drop both fields; re-create the `agent_run_cloud_has_no_local_assignment` check constraint without its `queue_position__isnull=True` term; reword the `WAITING_FOR_WORKTREE` comment.
- **Migration** `0025_drop_worktree_pool_columns`: `RemoveConstraint` → `RemoveField` × 2 → `AddConstraint` (constraint re-created without `queue_position`).
- **Serializers**: drop the key from the daemon-facing `runner/serializers.py` and the web-facing `app/serializers/issue.py`. `apps/web` has no references, so payloads tolerate the absence.
- **Null-only writers**: removed the `queue_position=None` writes in `run_lifecycle.py`, `run_endpoints.py` (accept), `agent_run_finalization.py`, `session_service.py`, `runs.py`, `utils/issue_move.py`; reworded the discard-branch comments in `sessions.py` (poll) and `matcher.py`.
- **Tests**: `test_drain_pod.py` keeps a heartbeat-ranking assertion without touching the dropped column; `test_worktree_pooling.py` updated so the queued-matrix, accept, cancel, and poll tests no longer read the dropped columns.

## Explicitly retained

- `AgentRunStatus.WAITING_FOR_WORKTREE` — historical rows still hold it and it keeps gating (busy/non-terminal status sets, work_item uniqueness).
- `RunQueuedEndpoint` — pre-retirement daemons still POST `run_id` + `queue_position`; the endpoint keeps returning a clean 200 and now records nothing from the body.
- The Rust `RunQueued` wire verb — out of scope, protocol slot retained for compatibility.

## Validation

- `grep -rn 'free_worktrees\|queue_position' apps/api/pi_dash` returns nothing outside historical migrations.
- Migration 0025 applies cleanly onto a fresh DB across the full chain; verified both columns are dropped and the constraint has no `queue_position` term.
- `pytest pi_dash/tests/unit/runner/ pi_dash/tests/unit/orchestration/` — 566 passed. (One unrelated pre-existing failure, `test_valid_agents_matches_runner_schema`, fails on `parents[6]` because the Rust `runner/` tree isn't in the api-only Docker mount — independent of this change.)

### DIFF
--- a/apps/api/pi_dash/app/serializers/issue.py
+++ b/apps/api/pi_dash/app/serializers/issue.py
@@ -1314,7 +1314,6 @@ class IssueDetailSerializer(IssueSerializer):
             "id": str(run.id),
             "status": run.status,
             "executor_kind": run.executor_kind,
-            "queue_position": run.queue_position,
             "runner": str(run.runner_id) if run.runner_id else None,
             "runner_name": run.runner.name if run.runner_id and run.runner else None,
             "created_at": self._serialize_datetime(run.created_at),
--- a/apps/api/pi_dash/runner/migrations/0025_drop_worktree_pool_columns.py
+++ b/apps/api/pi_dash/runner/migrations/0025_drop_worktree_pool_columns.py
@@ -1,0 +1,59 @@
+# Copyright (c) 2023-present Pi Dash Software, Inc. and contributors
+# SPDX-License-Identifier: AGPL-3.0-only
+# See the LICENSE file for details.
+
+"""Drop the two dead worktree-pool columns.
+
+The worktree pool is retired (PDASHOSS01-134 removed it from the daemon,
+PDASHOSS01-137 retired the cloud-side half). Two columns added by
+``0017_worktree_pooling`` outlived the retirement and have no live readers or
+writers:
+
+- ``Runner.free_worktrees`` — the per-runner capacity hint. No longer written
+  or read; the matcher ranks eligible idle runners by heartbeat alone.
+- ``AgentRun.queue_position`` — the display-only local-queue position. Every
+  remaining write only cleared it to ``None``.
+
+Dropping ``queue_position`` requires re-creating the
+``agent_run_cloud_has_no_local_assignment`` check constraint without its
+``queue_position__isnull=True`` term, so this migration pairs a
+Remove/AddConstraint around the field drop.
+"""
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("runner", "0024_managed_runner_provisioning"),
+    ]
+
+    operations = [
+        migrations.RemoveConstraint(
+            model_name="agentrun",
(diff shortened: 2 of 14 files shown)


irichard00:
## Code review — PDASHOSS01-156 (Drop dead worktree-pool columns)

**Verdict: approved — no defects found.** Reviewed the full diff against the issue scope and acceptance criteria.

### What I checked
- **Source removals** — every dropped write of `queue_position` / `free_worktrees` was a null-only writer (`run_lifecycle.py`, `run_endpoints.py`, `agent_run_finalization.py`, `session_service.py`, `runs.py`, `utils/issue_move.py`, `sessions.py` discard branch). Each `update_fields`/`update(...)` list is correctly pruned to match. Serializer keys dropped in `runner/serializers.py` (daemon-facing) and `app/serializers/issue.py` (web-facing); `apps/web` has zero references, so payloads tolerate the absence.
- **Migration 0025** — `RemoveConstraint → RemoveField ×2 → AddConstraint`. The re-created `agent_run_cloud_has_no_local_assignment` constraint carries **no** `queue_position` term, and its literal values (`["local_runner","managed_runner"]`, `"cloud_agent"`) exactly equal `list(MACHINE_EXECUTORS)` / `AgentExecutorKind.CLOUD_AGENT` in the model.
- **Migration drift** — `makemigrations runner --check` reports churn on this branch (constraint recreate + index renames + `executor_kind`/`id` alters). I verified the **identical** drift exists on `main` in a clean worktree: it's pre-existing Django 4.2.30 baseline churn, **not introduced by this change**. Django's would-be regenerated constraint contains no `queue_position`, confirming 0025 matches the model.
- **Acceptance** — `grep -rn 'free_worktrees\|queue_position' apps/api/pi_dash` (excluding migrations) is empty. Back-compat is preserved: the pre-retirement poll body and `/runs/<id>/queued` POST still 200 (`test_worktree_pooling.py`), and the tests assemble the legacy wire keys at runtime so the grep stays clean without losing coverage — faithful, since the endpoints ignore the body.
- **Retained per scope** — `WAITING_FOR_WORKTREE` status and `RunQueuedEndpoint` are untouched; the Rust `RunQueued` verb is out of scope.

No changes requested.

PR #388, opened 2026-09-14
PDASHOSS01-162 Verify the renamed agent engine's sandbox on macOS and Windows, and during a real agent run

irichard00:
## What

`prepare-agent.sh` ships the upstream Codex binary renamed to `pidash-agent-engine`. Codex re-execs helper copies of itself (`arg0` / `current_exe()`) to build the sandbox that walls in the commands an agent runs. If a renamed binary fails to self-locate, commands could run **without** the sandbox and nothing would say so — a silent safety-feature failure (PDASHOSS01-162, design §24.8, PRIVATEPI1-47 item 4).

This adds a CI regression guard so a future `CODEX_BUNDLE_VERSION` bump that breaks `arg0` dispatch fails the build instead of shipping an un-sandboxed engine.

## Changes

- **`desktop/tests/sandbox_rename_smoke.py`** — stages the pinned upstream engine under its shipped name (`pidash-agent-engine`) and runs the ticket's probe: a no-sandbox control write must succeed, and the same out-of-bounds write attempted through `pidash-agent-engine sandbox` must be refused. It reuses `bundle_agents.py`'s download + checksum + archive-member logic (one source of truth for the asset), and resolves the engine tag straight from `prepare-agent.sh` so the guard always tests the version the bundle actually ships. Both documented traps are handled: `CODEX_HOME` is kept out of the system temp dir (trap 1), and the check asserts the command actually executed — not merely that the probe file is absent (trap 2).
- **`.github/workflows/pull-request-test-desktop.yml`** — runs the guard in the existing `native` job (Linux). Needs no model, session, or credential; only the pinned-release download uses the network.

## Validation

- Guard **passes** against the pinned engine (renamed binary blocks the write). Full `--stage` path (download → checksum → rename → probe) exercised end-to-end.
- Negative test: a fake engine that runs the command un-sandboxed makes the guard **fail** (exit 1) — confirms it actually catches a broken sandbox.
- Existing `desktop/tests` packaging suite still green (11 tests); the smoke is not picked up by the `test_*.py` discovery (invoked explicitly).

## Scope note (verification ACs tracked on the issue)

This PR delivers the **CI regression** acceptance criterion. Manual macOS **Apple Silicon** verification was also run this session and passes (filesystem + network blocked, controls pass) — recorded on the issue. The remaining acceptance criteria (Intel Mac, Windows, the signed/notarized build, and a real managed agent run needing model creds) require hardware/credentials not available to the automated run and are called out on the issue for a human.


irichard00:
## Review — sandbox rename CI guard ✅

Reviewed the two-file diff. **Approving; no changes required.** The guard is correct, the cross-file API usage is real (not hallucinated), and — most importantly — the `cargo test + packaging tests` job (which now runs this step) **passed in GitHub Actions on `ubuntu-22.04`**, proving the renamed engine's Linux sandbox actually engages and blocks the out-of-bounds write on the real runner.

Verified during review:

- `bundle.assets(target)[1]` unpacks the `openai/codex` 4-tuple exactly as `bundle_agents.py` defines it; `download(repo, tag, asset)` and `executable_bytes(payload, archive, member)` signatures match. The guard reuses the bundler's download+checksum+member logic, so it can't disagree with the shipped bundle about which asset the engine comes from.
- `resolve_engine_tag()`'s regex actually parses `rust-v0.153.4` out of `prepare-agent.sh` (confirmed by running it), so the guard tracks the version the bundle ships and can't silently drift.
- Engine name `pidash-agent-engine` matches `ENGINE_BIN` in `managed_runner.rs`.
- The step runs only on the `ubuntu-22.04` `native` job, so `host_target()`'s Windows `SystemExit` is unreachable in CI; the `.exe` suffix path is still correctly handled for manual `--engine` use on Windows.
- Both documented traps are covered: CODEX_HOME kept out of `tempfile.gettempdir()` (with an explicit assertion), and the `PROBE_START` marker check prevents the "file absent ⇒ false pass" trap when the engine never ran the command. Negative case (an un-sandboxed engine) correctly yields exit 1.

Two optional nits, neither blocking:

1. The scratch tree under `~/.cache/pidash-sandbox-rename-smoke` is removed at start but not on success, so a `--stage` run leaves the downloaded engine binary behind. Harmless in ephemeral CI; minor local clutter.
2. The writability control is a Python `write_text`, not an engine-invoked no-sandbox `touch`. That proves the target dir is writable but is a hair weaker than the ticket's method (it doesn't prove the engine-spawned `sh` can write there). Adequate for ruling out a read-only target.

Scope note (not a PR defect): this PR closes only the CI-regression acceptance criterion. The remaining ACs — macOS Intel, the signed/notarized build, Windows, and a real managed agent run — are genuine hardware/credential blockers, already documented on the issue, and out of scope for this change.

PR #391, opened 2026-09-14
PDASHOSS01-159 Reconcile desktop agent reason-code copy with backend

irichard00:
Standalone reason-code fix for **PDASHOSS01-159** (the live bug the human flagged for its own PR; the four transport/storage/UI slices are separate and not in this PR).

## What was wrong
The desktop reason-code copy map (`AGENT_RUNTIME_REASON_MESSAGES` in `apps/web/ce/components/desktop/agent-runtime-edition.ts`) was missing entries for reason codes the backend actually emits, so those situations rendered a raw slug to the user instead of a sentence.

## Correction to the original framing
The issue thread assumed the map's `gateway_session_revoked` / `desktop_session_required` keys were typos of the `ManagedRunnerReason` enum values `gateway_scopes_missing` / `desktop_not_connected`, and that the fix was to **rename** them. That's not right, and a rename would have regressed working copy:

- `gateway_session_revoked` and `gateway_unavailable` are real literals emitted by `assistant/views/agent_profile.py::_classify_credential_error` (token endpoint, 401/503).
- `desktop_session_required` is a real literal from `managed_runner/permissions.py::IsDesktopSession`.
- `gateway_scopes_missing` / `desktop_not_connected` are **distinct** `ManagedRunnerReason` situations.

So the backend emits **two families** of codes to the desktop, not one. The fix is **additive**: keep the existing (correct) keys and add the missing ones.

## Change
- Add copy for `desktop_not_connected`, `gateway_scopes_missing`, `gateway_unavailable`; keep all existing keys.
- Add `apps/web/tests/desktop/agent-runtime-reason-codes.test.ts`: reads `ManagedRunnerReason` from `errors.py` plus the known non-enum desktop literals, and fails if the copy map drifts — a backend code with no copy, or a copy key matching no backend code. (`no_managed_runner_for_project` is excluded from the "needs copy" set because the desktop resolves it silently.)

## Validation
- `npx vitest run tests/desktop/agent-runtime-reason-codes.test.ts` → 2 passed.
- Verified the test genuinely fails on injected drift (removed a required key + added a bogus key → both assertions failed), then passes after restore.
- Pre-commit `oxfmt` + `oxlint --deny-warnings` clean.


irichard00 (commented):
Reviewed as the PDASHOSS01-159 review pass. No findings — this is sound to merge.

I verified the change against the backend rather than the plan, because the plan's original premise (that the two "wrong" keys were typos to be renamed) was itself wrong. What I confirmed:

- **Enum family** — `ManagedRunnerReason` in `apps/api/pi_dash/managed_runner/errors.py` declares exactly: `managed_runner_disabled`, `desktop_not_connected`, `llm_config_missing`, `gateway_scopes_missing`, `byok_not_supported_on_desktop`, `no_managed_runner_for_project`. The map covers the first five; `no_managed_runner_for_project` is correctly exempted (resolved silently by the desktop).
- **Non-enum family** — `gateway_session_revoked` and `gateway_unavailable` are returned by `_classify_credential_error` in `assistant/views/agent_profile.py:125-126`; `desktop_session_required` is the literal in `managed_runner/permissions.py:20`. `_classify_credential_error` returns *only* those two codes, so the family is fully enumerated.
- **Consumer** — `desktop-overlay/apps/web/core/services/agent-runtime.ts` routes both `response.data.error`/`.detail` (line 48) and `profile.reason_code` (line 111) through this map with a raw-slug fallback, so both families surface copy.

The additive fix (keep existing keys, add the three genuinely-missing ones) is right: a rename would have deleted live copy for `gateway_session_revoked`/`desktop_session_required`. Ran `agent-runtime-reason-codes.test.ts` on the branch → 2 passed.

One non-blocking note for later: the drift test reads the enum dynamically but guards the non-enum family against a hardcoded `NON_ENUM_DESKTOP_CODES` list. A *new* literal added in `agent_profile.py`/`permissions.py` wouldn't be caught as needing copy. That's an inherent limitation (those literals aren't machine-enumerable) and the list carries source pointers, so it's acceptable — just worth knowing the guarantee is weaker for that family.

PR #392, opened 2026-09-14
PDASHOSS01-159 (slice 3) Local SQLite chat history + concurrency guard

irichard00:
## PDASHOSS01-159 — slice 3: local SQLite chat history + concurrency guard

Third of the four slices (plus the standalone reason-code fix in #391) for direct local chat with the built-in agent engine. This slice is **storage only** and is independent of PDASHOSS01-158 — the transport and UI land in slices 2 and 4.

A local chat's transcript lives only on this machine, scoped per signed-in account, and is never relayed to or stored by the Pi Dash server.

### What's here (`desktop/src-tauri`)

- **`chat_history` module** — a per-account SQLite DB under the managed tree at `managed/chat/<account-key>/history.db` (directory `0700`, DB file `0600` on Unix, matching the existing credential handling). Schema: `sessions(id, title, workspace, project, working_dir, engine_version, engine_thread_id, created/updated)` and `events(id, session_id, seq, role, content, tool_calls, approval_decision, created_at)` with FK cascade and an ordering index, plus a `user_version` migration. A fresh connection per call keeps the module free of shared `!Sync` state; WAL + `busy_timeout` make overlapping UI calls safe.
- **Tauri commands** — create / list / get sessions, list / append events, `set_thread_id` (resume-by-thread-id), rename, delete-one, clear-all, and resolve the chat working directory (so the UI can show it in the header).
- **Concurrency guard (AC6)** — a local chat's working copy is resolved under `managed/chat/<account>/workdirs/…`, a *distinct* tree from managed runs' `managed/workdirs/…`, so a chat and a managed issue run can never modify the same path. The rule holds by construction and is asserted by a test.
- **Sign-out** — chat history is preserved by default (kept, hidden, until the same account signs in again — the approved behaviour). `managed_sign_out` gains an optional `clear_chat_history` request that wires the "clear on sign-out" opt-in; the existing no-arg overlay call is unchanged (a missing optional arg deserializes to `None`).
- **Account keys** are traversal-safe (filtered visible prefix + FNV-1a suffix over the raw bytes), so no account value can escape the chat tree and values that filter to the same prefix stay distinct.

### Design decisions (human-approved on the issue, 2026-09-14)

- Per-account scoping via an `account` identifier the overlay supplies (same pattern as the existing `workspace` arg). *Assumption, documented on the issue.*
- Separate chat working copy → the concurrency AC holds by construction.
- History kept on sign-out by default; explicit opt-in to clear.

### Validation

- `cargo test -p pi-dash-desktop` → 26 passed, 1 ignored (7 new `chat_history` tests: persist/reopen, event ordering + FK cascade, per-account isolation, clear-one-account, working-dir non-collision, traversal-safe keys, unknown-session rejection).
- `cargo clippy` clean for the new/edited code (two pre-existing warnings elsewhere left untouched).
- Not covered here: the live desktop UI, cross-platform verification (AC8), and the transport that actually drives these commands (slices 2 & 4).


irichard00:
Independent review of the PDASHOSS01-159 stack found defects owned by this slice (full write-up: https://github.com/The-AI-Republic/pi-dash/pull/404#issuecomment-5687053599):

- **HIGH** — `managed_sign_out` (managed_runner.rs L423) runs `clear_account_history(...)?` **before** removing the model token / credentials.toml / config.toml. If the history wipe fails (a busy file under the working copy — common on Windows), sign-out returns early and **credentials are never deleted**. Reorder credential removal first, or make the history clear best-effort.
- **LOW** — `chat_working_dir` / `resolve_chat_working_dir` (L485) create the account tree at umask (0755) and never `restrict_dir`; only `open()` chmods. `restrict_file` also covers only `history.db`, not the `-wal`/`-shm` sidecars. Multi-user-host exposure; the '0600' doc claim doesn't cover sidecars/working copy.

PR #345, opened 2026-08-30
PDASHOSS01-94 pidash CLI: support setting a parent issue on create and patch (--parent)

irichard00:
## What

Add issue-parentage management to the `pidash` agent CLI, using the same work-items `POST`/`PATCH` endpoint the MCP path already calls (no new endpoint).

- `pidash issue create --parent <PROJ-123|uuid>` — resolves a project-scoped identifier to a UUID via the existing `resolve_issue` helper (raw UUID passed through as-is, mirroring `--state`), then sends `"parent": "<uuid>"` in the create body.
- `pidash issue patch <id> --parent <PROJ-123|uuid>` — same resolution in the PATCH body; counts as a valid mutation for the "at least one of ..." guard.
- `pidash issue patch <id> --clear-parent` — sends `"parent": null` to detach. `--parent` and `--clear-parent` are mutually exclusive (clap `conflicts_with`).

## How

- Body construction is factored into pure `build_create_body` / `build_patch_body` helpers so the field contract (including `parent: null` for clear, and that a parent mutation satisfies the guard) is unit-testable without a network round-trip.
- New `resolve_parent_id` helper reuses `resolve_issue` / `looks_like_uuid`; empty input is a clean pre-API `EXIT_INVALID` error, a nonexistent identifier surfaces as the GET's not-found error.
- No client-side cycle detection — the server/serializer owns cross-workspace/cross-project validation and cycle checks; its errors surface as-is.
- Updates the CLI help text and the `pidash-cli` prompt section (`prompting/sections/pidash-cli.md`) so agents learn the new flags.

## Tests

`cargo test -p pidash` (26 `cli::issue` unit tests, incl. new parent-body construction + `--parent`/`--clear-parent` clap conflict) and `cargo clippy --all-targets -- -D warnings` both clean.

Related: PDASHOSS01-91, PDASHOSS01-92.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #402, opened 2026-09-15
PDASHOSS01-168 Agent prompt: finish a multi-part issue in one run instead of one PR per run

irichard00:
### Problem
An issue has a single shared run budget across In Progress / In Review / In Test. The coding-task prompt made the agent deliver at most one PR per run and advance the issue to In Review as soon as any PR opened. For a large-but-unambiguous issue this burned the budget: a split-gate round-trip before any code, then one part per run, plus review runs that could only bounce partial work back to In Progress (evidence: PDASHOSS01-159 — 7 runs, 5 PRs, 2 runs with no code).

No technical limit forces this — it comes entirely from the prompt sections.

### Changes (`apps/api/pi_dash/prompting/sections/`)

**`analyze-and-scope.md`** — the split gate now plans instead of always blocking:
- New decision outcome **Proceed with a multi-part plan** for a large-but-clear issue: record the parts / order / dependencies in the workpad `### Plan`, post a heads-up comment (not a question), and start building in the same run. No block.
- **Clarify** is now explicitly reserved for an open product/UX/architecture *decision* — not for size.
- **Propose a split** is reserved for genuinely separate deliverables that belong in different issues.
- Cost-comparison wording reworded to be about unclear *direction*, not size; added a line telling the agent not to keep slicing an already-recorded/approved plan unless a new decision surfaces. Worked examples updated (multi-part plan vs. genuinely-separate split).

**`implementation.md`** — keep going after a PR, advance only when done:
- Steps 5–6 are now a loop over the plan's parts within one run (stacked branches when a part depends on the previous one, otherwise branched from base), with explicit run-end conditions (all done / all remaining blocked / running low on room — finish the part in hand).
- After a part's PR opens with parts remaining, the agent continues in the same run; a partial implementation must **not** move the issue to In Review or yield `done`.
- The acceptance-criteria / How-to-Test hand-off is posted **once for the whole issue**, listing every PR, only when every part is built.
- Step 9 routes on plan state: parts remaining → stay In Progress, yield `progressed` (or `waiting_on_external` when only a merge is left); all parts built → In Review + `done`.

**`task-lifecycle.md`** (locked) — In Progress exits only when every planned part is built; added a paragraph that multi-part issues stay In Progress until the whole issue is done. No change to `pidash run yield` outcomes or the stage table's allowed moves beyond this condition.

### Tests
[...]

PR #403, opened 2026-09-15
PDASHOSS01-169 Agent splits oversized issues into child issues (task-level), not into per-run slices

irichard00:
## PDASHOSS01-169 — Split oversized issues into child issues (task-level)

When an issue is really **several genuinely separate tasks**, the agent now breaks it into **child issues** — each a self-contained task with its own In Progress → In Review → In Test lifecycle and its own ticking budget — instead of slicing the work across its own runs on one issue (which shares a single 10-run budget; evidence: PDASHOSS01-159).

### Parent flags and the collision with `main`
`main` shipped `--parent` on `issue create` / `issue patch` in **PDASHOSS01-94 (#345)** after this branch was cut, so both sides implemented the same two flags. The merge takes **main's implementation wholesale** — it is a superset (adds `--clear-parent` → `parent: null`, a `ParentPatch` enum, an empty-`--parent` guard, and clap conflict tests). This branch's duplicate create/patch parent tests are dropped in favour of main's.

What remains genuinely new here:

### CLI (runner)
- `pidash issue list --parent <ident>` lists a parent's children, so a later run finds existing children and never files duplicates. (`create`/`patch --parent` now come from `main`.)
- `build_list_params` takes the resolved parent UUID, keeping the query contract unit-testable without a network round-trip.

### API
- The work-item list endpoint honours a `parent` query param (filters by the parent's UUID); a malformed UUID returns 400, not 500. Added `PARENT_PARAMETER` to the OpenAPI docs.

### Prompt (`analyze-and-scope` step 6 + `workpad-setup`)
- New **"Split into child issues"** outcome: create children with `--parent`, list existing children first (idempotent across runs), guardrails (≤6 children, single-run sizing, depth 1), start unblocked children in In Progress, **park the parent in Todo**, comment the plan on the parent, and promote the parent to In Review when the children finish.
- **Split on independence, never on size** — a one-line test (could someone build child B without ever seeing child A's code or decisions?), explicit split / do-not-split lists, size as a fallback only, and "pin any shared contract first".
- The parent is parked in **Todo** rather than left In Progress reporting `waiting_on_external` — verified in `scheduling.py` that `waiting_on_external` is **not** a stopping outcome and keeps ticking for In Progress, so it would burn the parent's budget. Todo is non-ticking → the parent goes dormant. **No change to the locked ticking code.**
- The children are recorded in the **parent's description** (not only in a comment): a child's prompt inlines the parent's description but passes its comments as a count, so the description is the only signal by which a child can recognise a tracking parent.
- Parent-readiness (step 2) and base-branch resolution (`workpad-setup`) treat a branchless **tracking parent** as context, not a blocker: children base off the project base branch, or off a sibling child's branch when they depend on it.

### Tests (all run against the merged head)
- `cargo test --lib` — **489 passed**, incl. `build_list_params_includes_resolved_parent` and main's `--parent` / `--clear-parent` tests.
- Prompting unit tests — **143 passed** (render tests for the new prompt branch).
- `test_issue_list_parent_filter.py` — **3 passed** (children-only filter, unfiltered list, 400-on-bad-UUID).
- Full Python unit suite — same 3 pre-existing failures as pristine `main`, no regression.


irichard00:
## Review — PDASHOSS01-169 (In Review pass)

Reviewed the full diff against the `pi-dash/pdashoss01-168` base (9 files, +432/−59). **No blocking findings** — the change is correct, well-factored, and each acceptance criterion is covered by a test.

**What I checked**
- **CLI (`runner/src/cli/issue.rs`)** — `--parent` on create/patch/list. Body/param assembly refactored into pure `build_create_body` / `build_patch_body` / `build_list_params`; `resolve_parent_id` correctly passes a UUID through and resolves a `PROJ-N` identifier otherwise. The empty-body guard message now includes `--parent`. Unit tests exercise all three builders (present/absent parent, parent-only patch, query-string shape).
- **API (`api/views/issue.py`)** — the `parent` query param filters both `issue_queryset` and `total_issue_queryset` by `parent_id`; malformed UUID → 400 (not 500). `uuid` is imported; the base list queryset does **not** exclude sub-issues, so the filter returns children as intended. `PARENT_PARAMETER` is defined and re-exported from `utils/openapi/__init__.py` (the re-export is easy to miss and is covered).
- **Prompt** — the new "Split into child issues" outcome in `analyze-and-scope.md` step 6 creates children itself (`--parent`), lists existing children first for idempotency, states guardrails (≤6, single-run, depth 1), starts unblocked children, parks the parent in Todo (not `waiting_on_external`, which keeps ticking), and promotes to In Review when the last child finishes. The tracking-parent readiness case and the `workpad-setup.md` base-branch case both treat a branchless tracking parent as context, not the "parent in progress with no branch" blocker.
- **Tests** — render tests assert the new prompt branch (I hand-verified the string assertions match the source), API contract test covers the parent filter + 400 case, Rust tests cover `--parent`.

**Two coordination notes (not code defects, carried to In Test):**
1. This PR is based on `pi-dash/pdashoss01-168` (#402) because both edit `analyze-and-scope.md` step 6. **#402 must land first**, then this PR's base retargets to `main` (or is retargeted manually).
2. The installed runner (`pidash` 0.1.23) won't have `--parent` at runtime until it's rebuilt from this branch; the flags are unit-tested against source.

Approving the review — moving the issue to In Test.

irichard00:
Pushed `688c3aba`: **the split gate now keys on independence, not on size or layer.**

**Why.** The previous wording split when the work was "clearly more than one run's work" or "span[s] different areas (runner protocol vs desktop host vs web UI)".
- The first is a size test. There is no run timeout or turn cap, and a modern agent can finish a large, clear change in one long run, so the agent had nothing concrete to measure against.
- The second is a layer test, and it is what broke voice dictation (PDASHOSS01-148). Backend and frontend were sibling issues because they were "different areas". Each built against its own guess of the upload contract, every sibling went green, and the mic shipped broken (fixed in #424).

**What changed in `analyze-and-scope.md`**
- One-line test: *could someone build child B correctly without ever seeing child A's code or decisions?* If not, do not split.
- Explicit **Split when** list: independently valuable and verifiable; different repo or separate deploy; different human decision or owner; risky enough for its own review (migration, security).
- Explicit **Do not split when** list: shared interface not yet fixed; parts cannot be verified alone; a reviewer must read them together.
- **Size is only a fallback**, for a real limit such as context, and the agent must say so.
- **Count the cost**: every child pays its own build, review and test runs.
- **Pin any shared contract first**: if a split must cross an interface, write the contract on the parent before creating children, and land the producer before the consumer starts.
- The example comment now shows a split that follows these rules.

Parts of one feature still go to the multi-part outcome from #402, unchanged.

**Tests.** New `test_coding_task_split_standard_is_independence_not_size` pins the standard and asserts the old size and layer phrases are gone. Prompting unit suite: 129 passed. One unrelated test (`test_context_code_reviews_excludes_soft_deleted`) fails only in my local throwaway setup because it had no message queue; it fails identically without this commit.

irichard00:
## Review — PDASHOSS01-169, second pass (the **merged** diff, base `main`)

The earlier review on this PR read the *pre-merge* diff stacked on `pi-dash/pdashoss01-168`. Since then the branch merged `origin/main` and took main's `--parent` implementation for `issue create` / `issue patch`, so the create/patch code in this PR is no longer what that review looked at. This pass re-reads the whole diff as it now stands (11 files, +417/−27, merge-base is exactly `origin/main`).

**Result: approved**, with one defect found and fixed in this pass (`5600fb0e`), plus two non-blocking notes.

### Defect found and fixed — the guardrail contradicted the tracking-parent mechanism

The split outcome tells a run to rewrite the parent's description with the child list, and that list is load-bearing: `prompting/context.py` inlines `parent.description` into a child's prompt but passes comments as a **count**, not content. The description is the only place a child can read its siblings from. That's why `911a6c44` added it.

But two sections still said, unqualified, that the agent must not do it:

- `sections/guardrails.md:8` — *"Do not edit the issue title or description for planning or progress tracking."* This section is `customizable: locked`, i.e. the hard-rules block.
- `sections/pidash-cli.md:33`, under `issue patch` — *"Do **not** edit title or description for planning or progress tracking."*

The new step-6 text carves the exception out in place (*"This is the one place you do edit a description"*), so a careful reader reconciles it — but an agent that treats the locked Guardrails section as authoritative skips the description rewrite, and the tracking-parent signal is never written. That is exactly the failure `911a6c44` exists to prevent, reintroduced from the other end.

Fixed by qualifying both lines with the single exception and pinning it with a render test (`test_coding_task_guardrails_allow_the_tracking_parent_child_list`). Prompting unit suite: **144 passed** (was 143).

### Verified, no findings

- **API `parent` filter** (`views/issue.py`) — `uuid` is imported at module scope; a malformed value returns 400 before any queryset work; the filter is applied to **both** `issue_queryset` and `total_issue_queryset`, so the paginated count matches the page. `Issue.issue_objects`' manager excludes triage/archived/draft only — it does **not** exclude sub-issues, so children are genuinely returned. `PARENT_PARAMETER` is defined in `utils/openapi/parameters.py` *and* re-exported from the package `__init__`, which `views/issue.py` imports from — the omission that would raise `ImportError` at URLconf load is not present.
- **CLI** — `build_list_params` is a pure function returning `Vec<(&'static str, String)>`, tested for both the present and absent cases including the assembled query string. `cmd_list` resolves `--parent` through main's `resolve_parent_id`, which rejects an empty value and passes a UUID through unresolved — so `--parent ENG-1` and `--parent <uuid>` both land on the `parent` query key as a UUID, which is what the server filters on.
- **Prompt** — the split standard is independence-first with an applicable one-line test, explicit split / do-not-split lists, size as a fallback only, and a run-cost count. Idempotency (list children first), guardrails (≤6 / single-run / depth 1), starting unblocked children, parking the parent in **Todo** rather than `waiting_on_external`, cutting no branch on the parent, and last-child promotion are all present. `{{ project.identifier }}` and `{{ issue.identifier }}` both resolve in the `coding-task` context (`context.py` builds a `project` dict), so the commands render concrete.
- **Boundary with #168** — `Proceed with a multi-part plan` is still the home for one task needing several PRs; the two outcomes read as distinct alternatives, not overlapping ones.
- **Cloud executor** — `CLOUD_RECIPES` shares no local-runner section, so none of this CLI guidance leaks into Cloud prompts. Correct by design, nothing to do.

### Non-blocking notes

1. **`prompting/fragments/08_analyze_and_scope.md` still carries the old text** — *"Propose a split … **Do not create child issues yourself** — leave triage to the human."* That directory only seeds the legacy `PromptTemplate` row; live prompts are composed from `sections/` via `compose()` / `build_first_turn`, so this does not reach a run. Recent prompt work (PDASHOSS01-93, the ticking series, #345) has likewise left `fragments/` untouched, so this looks like the established direction rather than an oversight — flagging it only because PDASHOSS01-160 did update both.
2. **A parked parent has no safety net.** Todo is non-ticking, which is the point — but it also means the parent is promoted to In Review *only* by a child's agent run. If a human closes the last child by hand, or the last child's run doesn't notice it is last, the parent sits in Todo indefinitely with nothing scheduled to re-check it. The ticket explicitly sanctioned reusing Todo, and the alternative needs a ticking-code change this PR deliberately avoids, so this is the accepted trade-off — worth knowing before it happens in production rather than after.

Moving the issue to **In Test**.

PR #401, opened 2026-09-15
PDASHOSS01-159 slice 2: desktop-host local-chat transport

irichard00:
Slice 2 (the desktop-host local-chat transport) — stacked on slice 1b (#400), which stacks on slice 1a (#398) and the `pidash-ipc` crate (#395). Rebases onto `main` as the stack merges down.

This is the desktop-host half of direct local chat: the webview talks to the bundled engine through the Tauri host and the daemon's IPC socket, never through the Pi Dash cloud chat relay. It consumes the slice-1a wire protocol and the slice-1b daemon handler.

**Transport surface (`desktop/src-tauri/src/chat.rs`):**
- **UI → engine** — Tauri `invoke` commands `chat_warm`, `chat_send`, `chat_cancel`, `chat_close`, `chat_decide`. `warm`/`send` open a streaming connection; `cancel`/`close`/`decide` ride a *second* connection because the streaming one is busy pumping the turn — matching slice 1b's out-of-band approval design.
- **engine → UI** — every streamed `Response::Chat*` frame is re-emitted to the webview as a Tauri event `chat://frame` (payload = the frame with its usual `result`/`data` tagging). Transport failures the webview can't otherwise see (daemon not running, dropped connection) surface as `chat://error`, keyed by `chat_session_id`. This is the event contract the desktop-overlay local `ChatTransport` (slice-4 remainder) subscribes to.
- A per-session registry lets `chat_close` abort the desktop-side reader task after the daemon tears the session down.

**Why design-wise:** `Client::read_next()` (from #395) already supports subscribe-style streaming, so the transport is a thin pump: `call()` sends the request + reads the first frame (so a synchronous `Error` — e.g. a pre-slice-1b `501`, or a connection refusal — is surfaced immediately), then `read_next()` streams the rest until the terminal `Ack`. No second IPC client is introduced.

**Deliberately out of scope here:**
- Local SQLite history (slice 3, #392) is on a separate branch off `main`; the transport neither reads nor writes history. Persistence is wired where the two integrate.
- The desktop-overlay local `ChatTransport` impl + the local-labelled built-in-agent entry in `RunnersSideNav` (slice-4 remainder) — they subscribe to the `chat://frame`/`chat://error` events this slice emits.

Whole-issue AC 1–8 are still not all met, so the issue stays In Progress. The criteria below scope **this slice**.

### Acceptance Criteria
[...]

PR #400, opened 2026-09-15
PDASHOSS01-159 slice 1b: daemon-side local-chat IPC handler + AC6 guard

irichard00:
## Slice 1b — daemon-side local-chat IPC handler + AC6 concurrency guard

Part of **PDASHOSS01-159** (Desktop: direct local chat with the built-in agent engine). This is the daemon-side half of slice 1: it replaces the `Chat{Warm,Send,Cancel,Close,Decide}` **501 stub** (landed in slice 1a, #398) with a real handler that drives the built-in engine directly over the daemon's local IPC socket and streams the turn's events back on the same connection — **never through the Pi Dash cloud relay**.

Stacked on slice 1a (**#398**), which is stacked on the shared `pidash-ipc` crate (**#395**, PDASHOSS01-158). Base is set to the slice-1a branch so this diff is only slice 1b; it rebases onto `main` as the stack merges.

### What's here

- **`runner/src/ipc/chat.rs`** (new) — a **self-contained parallel lane**, deliberately *not* a reuse of the cloud `ChatWorker`. It drives `AgentBridge` itself and writes `Response::Chat*` IPC frames, leaving the cloud chat path **completely untouched** (issue AC7). The `BridgeEvent` → `Response::Chat*` translation mirrors the cloud `ClientMsg::Chat*` translation field-for-field, minus the cloud-only `chat_timing` telemetry.
  - `ChatWarm` → spawn/reuse + warm the engine, stream `chat_warmed`.
  - `ChatSend` → run a turn, stream `ChatMessageStarted` / `ChatEvent` (assistant deltas + raw) / `ChatApprovalRequest`, return terminal `ChatMessageCompleted` / `ChatFailed`.
  - `ChatCancel` → trips a per-session signal the active turn observes and interrupts.
  - `ChatClose` → interrupt, tear the engine down, drop the session, emit `ChatClosed`.
  - `ChatDecide` → routes into the shared `ApprovalRouter` as `DecisionSource::Local`, unblocking a turn parked on an approval (answered out-of-band on a second connection).
  - Sessions persist across requests/connections in a per-runner `ChatRegistry` so a Codex thread survives between turns (the concrete Codex bridge has no resume-by-id).

[...]

PR #404, opened 2026-09-15
PDASHOSS01-159 Desktop: direct local chat with the built-in agent engine (no cloud relay, local chat history)

irichard00:
## PDASHOSS01-159 slice-4 remainder — desktop local chat transport + picker entry

The final piece of direct desktop↔local-engine chat: it wires the shared runner
chat UI to the bundled agent engine over local Tauri IPC, never the Pi Dash
cloud chat relay.

### Base & merges
- **Targets `pi-dash/pdashoss01-159-slice2-desktop-transport` (#401)**, which already stacks #395 → #398 → #400 → #401.
- **Also merges** `pi-dash/pdashoss01-159-slice3-chat-history` (#392, local SQLite history) and `pi-dash/pdashoss01-159-slice4-chat-transport` (#393, the `ChatTransport` seam) via merge commits, so both stay reviewable on their own. Conflicts (`Cargo.toml`/`Cargo.lock`/`main.rs` — combining the `chat` and `chat_history` modules + `uuid`/`rusqlite` deps) were resolved in this branch only; nothing was pushed back to #392/#393.

### What this adds
- `desktop-overlay/.../services/local-chat-transport.ts` — a `ChatTransport` backed by Tauri IPC. Verbs call the slice-2 commands (`chat_warm`/`chat_send`/`chat_cancel`/`chat_close`/`chat_decide`) and slice-3 history commands; `subscribeChatEvents` listens on `chat://frame`/`chat://error` and translates each daemon `Response::Chat*` frame into the server-shaped `IAgentChatEvent` the UI already renders (assistant-delta grouping, approval, completion/failure/close). Each turn is persisted to local history as it streams. Registered desktop-only at boot via `setChatTransport`, with a runner-detail shim for the built-in agent id.
- Runner picker: a `useLocalChatContacts` seam (empty in the shared cloud web app) overridden in the desktop build to list the built-in agent, labelled **Local**.
- Chat header shows the session working directory when present.

### Verified here
- `cargo test` (desktop crate): 32 passed, 1 ignored (pre-existing engine smoke test).
- Overlay suite (`test-overlay.sh`, the merged tree the app bundles): 38 passed, incl. 11 new `local-chat-transport` tests.
- `apps/web` cloud chat/runner suites: 55 passed (unchanged); `tsc`: only the 3 pre-existing `agent-status.tsx` errors; oxlint `--deny-warnings` + oxfmt clean.

### Needs a real desktop run (no bundled engine in CI)
Streaming against a live engine; a network capture showing no chat traffic to the Pi Dash API; history surviving a restart; and the macOS/Windows/Linux three-platform check.


irichard00:
## Review — slice-4 remainder (local `ChatTransport` + picker entry + working-dir header)

Reviewed the integration slice that closes out PDASHOSS01-159. This was the one PR in the 8-PR set that hadn't yet had its own review pass (#391–#401 each did). **No blocking findings** — the code is clean, well-documented, and the trade-offs are called out at the source.

What I verified:

- **Cloud path is untouched (AC7).** The picker seam `useLocalChatContacts()` returns `[]` in `apps/web`; the built-in "Local" entry only appears via the `desktop-overlay/` override, and that override is itself guarded by `isDesktop()`. The shared cloud web app cannot render the built-in agent by construction.
- **Frame → `IAgentChatEvent` mapping** in `ChatFrameTranslator` covers all seven `Response::Chat*` frames, groups `assistant_delta`s onto the active `message_id`, and emits a monotonic `seq` the page dedupes on. `chat://frame`/`chat://error` are filtered by chat-session id before dispatch.
- **Persist-as-you-stream** writes the user turn on send, the engine thread id on `chat_started` (resume), and the assistant message on `chat_message_completed` — matching "the app DB is the source of truth."
- **Registration** (`registerLocalChatTransport`) is desktop-only and idempotent (guards against chaining the runner-detail fetcher onto itself).

Non-blocking observations (all already intentional per the code comments — noting them so a human reviewer has them, not asking for changes):

1. `subscribeChatEvents` ignores the `_after` cursor and starts a fresh translator at `seq = 0` on every (re)subscribe. Fine because live frames only arrive for in-flight turns and historical messages come through `listChatMessages`, but worth a glance if resume-mid-turn ever misbehaves.
2. Streamed `assistant_delta` text is not persisted incrementally — only the final assistant message on `chat_message_completed`. A crash mid-turn loses partial assistant text (the user turn survives). Acceptable given the stated source-of-truth model.
3. Runner id rides in the slice-3 `project` column and session `status` is always `"open"` (slice-3 has no runner/status column). Documented; a closed local session stays readable rather than marked closed.

Automated verification the author ran reproduces here as described (desktop `cargo test` 32/1-ignored, overlay suite 38 incl. 11 new transport tests, apps/web 55 unchanged, tsc only pre-existing errors, oxlint/oxfmt clean). What remains is genuinely out of an agent sandbox's reach: live streaming/approvals/interrupt against a bundled engine, the no-relay network capture, history surviving a real restart, and the macOS/Windows/Linux check.

irichard00:
## Independent review — PDASHOSS01-159 local-chat stack

This is a fresh, independent pass over the whole feature (all prior reviews on this issue were the author reviewing its own diff). I read the actual source on `pi-dash/pdashoss01-159-slice4-remainder` and verified each finding against the code (file:line + a concrete failure). Grouped by the PR that owns the fix. **Net: real defects across all four code slices — several serious. Recommend addressing before merge.**

---

### HIGH

**H1 — An abandoned approval permanently wedges the runner** · `runner/src/ipc/chat.rs` `resolve_approval` (~L693) · owner **#400**
`resolve_approval` awaits `rx.recv()` in a loop with **no timeout and no cancel arm** (`expires_at` is advisory — `ApprovalRouter::expire()` is never called, and the loop treats `Expired` as `continue`). While parked it holds the `LocalTurnGuard` and the session `rt` lock.
Failure: a `ChatSend` raises an approval the policy won't auto-decide → `ChatApprovalRequest` emitted → the desktop never sends `ChatDecide` (tab closed / host crash / socket drop). Then, until daemon restart: `ChatCancel` is lost (the task is parked inside `resolve_approval`, not in `drive_turn`'s `select!`, so the pinned `cancelled` future is never polled); `ChatClose` hangs on `session.rt.lock()`; and because `chat_active` stays `true`, **every** subsequent managed `Assign` is NACK'd (supervisor.rs). One unanswered prompt kills the runner for both chat and managed work. Breaks AC1 (interrupt) and regresses AC7 (existing runner unusable). Fix needs a design call: a cancel arm on the recv + a real TTL/expiry.

**H2 — A completed assistant reply can vanish from the UI *and* from history** · `desktop-overlay/apps/web/core/services/local-chat-transport.ts:488` + `apps/web/app/(all)/[workspaceSlug]/runners/chat/[runnerId]/page.tsx:258` · owner **#404**
Assistant text is persisted only when `frame.data.assistant_message` is truthy (transport L488). The daemon fills it from `assistant_text_from_done_payload` (chat.rs L779), which returns `None` unless the engine's *done* payload is a string or an object with one of six **guessed** keys (`result/message/text/output/summary/content`). On `chat_message_completed` the transport also emits `turn_completed`, and the page's `useEffect(() => setLiveMessages(messages ?? []), [messages])` (L258-259) unconditionally **replaces** the streamed reply with the refetched DB list — which has no assistant row. There is no fallback to the accumulated `assistant_delta` text (that's the only place the full reply exists in this case).
Failure: engine streams "Hello" via deltas, then completes with a done payload that doesn't match those keys → nothing persisted → refetch wipes the streamed reply. After restart the user sees their question with no answer (the user turn *is* persisted at L355; only the assistant turn is lost). Breaks AC1/AC4. Whether it fires depends on the real engine's done-payload shape — **this must be checked against a live engine**, and a delta fallback added regardless.

**H3 — Sign-out can leave engine/machine credentials on disk** · `desktop/src-tauri/src/managed_runner.rs:423` · owner **#392**
`managed_sign_out` runs `clear_account_history(...)?` (propagating errors) **before** removing `model_token_file`, `credentials.toml`, `config.toml`. Those credential removals are best-effort, but the history clear is gated by `?`.
Failure: user opts into clear-on-sign-out; a file under the chat working copy is held open (engine, a tool, a `.git/index.lock`, or any open handle on Windows) → `remove_dir_all` errors → `managed_sign_out` returns early → **credentials are never deleted**. The user believes they signed out but the model credential + machine token remain and the bundled binaries stay usable. The least-critical step (deleting the user's own chat) gates the security-critical one. Reorder credential removal first, or make the history clear best-effort.

---

### MEDIUM

**M1 — `chat_close` reports an error on every successful close** · `desktop/src-tauri/src/chat.rs:162` (`call_once`) · owner **#401**
Daemon `handle_close` emits `Response::ChatClosed` via the sink **then** returns `Ok(Response::Ack)` → two lines on the wire (`ChatClosed\n`, `Ack\n`). But `chat_close` uses `call_once`, which reads only the first line and accepts only `Ack`; `ChatClosed` hits the `other => Err(...)` arm. Teardown actually succeeds (the reader is aborted regardless), but the webview gets `Err("unexpected response: ChatClosed …")` on every close.

**M2 — Send reader task + IPC connection leak on every turn** · `desktop/src-tauri/src/chat.rs:105-125` (`drive_stream`) & `:188` (`insert`) · owner **#401**
`drive_stream` ends only on `Ack` or EOF, but `handle_send` terminates a turn with `ChatMessageCompleted`/`ChatFailed` (never `Ack`) and `handle_conn` keeps the connection open for the next request. So after the terminal frame, `read_next()` blocks **forever**. Compounding it, `streams…insert(id, handle)` discards the previously-registered `JoinHandle` (drop detaches, does not abort), so `chat_close` can only reap the newest task. An N-turn session accumulates N-1 hung reader tasks + open IPC connections that only process exit reclaims. (The stub test appends a trailing `Ack` the real daemon never sends for a send, which is why this passed CI.)

**M3 — cwd confinement bypass via `..`** · `runner/src/ipc/chat.rs:730` (`resolve_chat_workspace`) · owner **#400**
The guard is `requested.starts_with(&workspace_path)` on a **non-canonicalized** path; `Path::starts_with` is lexical and `join` doesn't normalize. `cwd = "../../../../etc"` with `working_dir = /home/u/ws` → `/home/u/ws/../../../../etc`, whose components start with `[/, home, u, ws, …]`, so the check passes and the un-normalized path is used as the process cwd, resolving outside the workspace. Defeats the confinement the function documents and can point a chat at a managed run's working copy (AC6 separation). The test only covers absolute `/etc` and a plain `sub`, never `..`. Canonicalize (or reject `..` components) before the check.

**M4 — Mid-stream daemon disconnect is swallowed as success** · `desktop/src-tauri/src/chat.rs:122` · owner **#401**
`drive_stream` returns `Ok(())` on premature `None`/EOF (daemon crash mid-turn), so `spawn_stream` emits no `chat://error` and the UI's last frame was a non-terminal event → the spinner hangs, contradicting the module's documented "connection dropped → `chat://error`" contract. (A clean warm EOF is legitimately `Ok`, so this needs a "terminal-frame-seen" check, not a blanket flip.)

---

### LOW / notes

**L1 — AC6 double-acquire (TOCTOU)** · `runner/src/ipc/chat.rs:168` + `runner/src/daemon/supervisor.rs:1078` · owner **#400**
The assign lane reads `local_chat_active` **before** it stamps `rx_in_flight` (set_current_run, ~L1140); the chat lane's post-CAS re-read of `rx_in_flight` can't observe an assign that's between its own check and its stamp, so both lanes can acquire the same working copy. This is a genuine logical race against AC6 ("never … at the same time"). **Already documented by the author** as a known best-effort limitation with a hard-lock deferred — flagging it so the AC6 sign-off is honest, not as a surprise.

**L2 — Chat working-dir / WAL sidecars can be world-readable** · `desktop/src-tauri/src/chat_history.rs:485` (`resolve_chat_working_dir`) & `:419` (`open`) · owner **#392**
`restrict_dir` (0700) is applied only inside `open()`; the `chat_working_dir` command calls `resolve_chat_working_dir`, which `create_dir_all`s `chat/<key>/workdirs/…` at umask (0755) and never chmods. If the UI shows the working dir before a session is created, the account tree (and checked-out repo) is world-traversable on a multi-user host. Separately, `restrict_file` covers only `history.db`, not the `-wal`/`-shm` sidecars (0644) that hold the freshest, not-yet-checkpointed chat content. Doc comment claims "The DB file is 0600" — true for the file, not the sidecars/working copy.

**Notes (not blocking):** `events` has no `UNIQUE(session_id, seq)` — ordering is incidental on SQLite's single-writer serialization; a concurrent append errors rather than corrupts, but the constraint would make the guarantee explicit. And the chat path validator (`safe_component`, ASCII-only) diverges from the managed one (`valid_component`, Unicode) — not an escape (both block `.`/`/`/`..`), but a Unicode workspace slug that works for managed runs fails for chat.

---

### Checked and clean (negatives considered)
Session-id filtering (frame + error) is consistent; subscription teardown is race-safe; per-account isolation and the account-key traversal guard (`[A-Za-z0-9_-]{0,48}-{16hex}`) hold; clear-all scopes to one account; cascade delete leaves no orphans; the AC7 cloud seam is inert in `apps/web` (`useLocalChatContacts()` → `[]`) and the "Local" entry is `isDesktop()`-gated; the happy/failed/cancelled terminal-frame paths in `drive_turn` are sound (UI hangs only via the approval-park path H1). The user message is persisted on send.

*(Independent multi-surface review pass. 🤖 Generated with [Claude Code](https://claude.com/claude-code))*

irichard00:
Follow-up: fixed 3 of the review findings on this integration branch (commit 43763082), each with a test:

- **M3** `runner/src/ipc/chat.rs::resolve_chat_workspace` now rejects any `..` component (a relative `../escape` previously passed the lexical `starts_with` check and resolved outside the workspace).
- **M1** `desktop/src-tauri/src/chat.rs::call_once` now treats `Response::ChatClosed` as a successful close — `handle_close` emits it before the terminal `Ack`, so `chat_close` was returning Err on every success. +`call_once_accepts_chat_closed`.
- **H3** `desktop/src-tauri/src/managed_runner.rs::managed_sign_out` holds the history-clear result until after credential teardown, so a wipe failure no longer leaves the model token + credentials on disk.

`cargo test -p pidash --lib chat` and desktop `cargo test` (33 passed / 1 ignored) are green.

Still open from the review (see the issue thread + workpad): **H1** approval-park (needs a cancel/timeout arm — design call + live engine), **H2** assistant-message loss (needs the live payload shape + delta fallback), **M2/M4** send reader-leak / mid-stream EOF (hinge on the real daemon connection lifecycle), **L1/L2**. These were deliberately left for a run that can verify against a bundled engine rather than patched blind on the final budgeted run.

Note: these three fixes are on #404 only; owning PRs #400/#401/#392 still show the original code — cherry-pick back or merge #404 last.

irichard00:
## Fixes for the independent-review defects (run: this tick)

Follows up the independent review (the H1–M4 + L1/L2 write-up above). The prior run landed **M1, M3, H3** (`43763082`). This run lands the remaining tangled streaming/approval defects plus L2, each with a test. Both Rust workspaces + the overlay suite are green.

Commits: `ccf0a0e1` (H1/H2/M2/M4), `763684a3` (L2).

### H1 — abandoned approval no longer wedges the runner
`runner/src/ipc/chat.rs::resolve_approval` waited on `rx.recv()` with **no cancel arm and no timeout**, so an unanswered approval (tab closed / host crash / socket drop) parked the turn forever and left `chat_active=true`, NACKing every future managed `Assign`. The wait now `select!`s over three arms, reusing the turn's single pinned `cancelled` future so no interrupt is missed:
- **cancel** (`ChatCancel`/`ChatClose`) → interrupt the engine, `expire` the approval, end the turn `cancelled`;
- **TTL backstop** (`APPROVAL_TTL_MINUTES`) → same teardown, terminal `ChatFailed{code:"approval_timeout"}`, for the silent-disconnect case a cancel can't cover;
- **decision** → unchanged.

Test: `abandoned_approval_is_interrupted_by_cancel` (drives a real fake-engine turn that requests an approval and hangs; a cancel unwinds it as `cancelled` and it does not park). The TTL arm is verified by inspection (a 10-min real timer isn't unit-friendly; the cancel arm exercises the same teardown path).

### H2 — assistant reply can no longer vanish
`local-chat-transport.ts` persisted the assistant turn only when the completed frame carried `assistant_message`, which the daemon fills only from guessed done-payload keys; otherwise the streamed reply was never written, and the page's refetch-on-`turn_completed` then wiped it from the UI too. Now the transport **accumulates the `assistant_delta` text** and persists it as a fallback, and persists **before** dispatching `turn_completed` so the refetch sees it in history. This is defensive regardless of the live engine's exact done-payload shape. Tests: delta-fallback persistence + accumulator reset between turns. (The AC7 cloud page is untouched.)

### M2 — send reader/connection no longer leaks
A `ChatSend` ends with `ChatMessageCompleted`/`ChatFailed` and the daemon holds the connection open with **no trailing `Ack`**, so `drive_stream` blocked on `read_next` forever, leaking a task + connection every turn. `drive_stream` now treats the terminal turn frames as end-of-stream. Test: `send_stream_returns_on_completed_without_trailing_ack` (server holds the socket open; the reader must still return).

### M4 — mid-stream EOF is surfaced
A daemon disconnect before any terminal frame was swallowed as `Ok`, hanging the UI. It's now an `Err` → `chat://error`. Test: `premature_eof_is_a_transport_error`. (The existing test that modeled a nonexistent post-completed `Ack` was corrected to match the real wire contract.)

### L2 — history files locked to the owner
`resolve_chat_working_dir` now restricts the account root to `0700` (the UI resolves the working dir for the header before the DB opens), and `open` chmods the WAL/SHM sidecars to `0600`. Test: `history_files_and_working_dirs_are_owner_only` (unix).

### Left open (intentional)
- **L1** (AC6 assign-vs-chat TOCTOU double-acquire) — kept as the author-documented best-effort exclusion; a true hard lock touches the managed-assign hot path and carries more risk than the sub-microsecond window warrants. Flagged so the AC6 sign-off stays honest.
- Live-desktop / 3-platform verification (bundled engine, network capture, restart, 0600) remains the human step.

### Validation run here
- `cargo test -p pidash` → 462 lib pass, 0 failed; new `abandoned_approval_is_interrupted_by_cancel`.
- `cd desktop/src-tauri && cargo test` → 36 pass, 1 ignored (pre-existing engine smoke); new chat + chat_history tests.
- `bash desktop/scripts/test-overlay.sh` → 40 pass (13 in `local-chat-transport.test.ts`, +2).
- clippy clean on the touched modules (only the known pre-existing warnings remain); oxlint/oxfmt clean on the overlay files.

These corrections live on the integration branch `#404` (the one branch that builds/tests all slices together); whoever merges the stack should carry them back to `#400`/`#401`/`#392`, or merge `#404` as the final state.

PR #406, opened 2026-09-16
PDASHOSS01-86 Support image uploads in Pi Dash CLI and MCP

irichard00:
## What & why

Issue **PDASHOSS01-86** asked for image-upload support in the Pi Dash CLI/MCP. Storage (S3/MinIO + `FileAsset`) already existed, but there was no way to attach an image to a comment from the CLI — it was text-only. This adds that path.

## Changes

**Runner CLI**
- `pidash comment add` gains a repeatable `--image <path>` flag. Body is now optional when at least one image is supplied (still requires body *or* image).
- Per image the runner does the same 3-step flow the web client uses: create the asset (`POST /api/v1/workspaces/{slug}/assets/`) → push bytes to the returned S3/MinIO presigned POST → mark uploaded (`PATCH .../assets/{id}/`), then embeds the editor's `<image-component src="{asset_id}">` node in the comment HTML.
- New `ApiClient::post_multipart` helper; enabled reqwest's `multipart` feature. MIME is inferred from the file extension (images only).

**API**
- `S3Storage.__init__` now accepts `is_server`. The API-key asset endpoints already called `S3Storage(request=..., is_server=True)`, but the parameter never existed — every call raised `TypeError`, so the whole `/api/v1/.../assets/` surface was dead. This fixes that and makes `is_server` select the MinIO endpoint (internal for server-to-server, the public request host otherwise).
- `GenericAssetEndpoint` GET/POST now sign against the public host so the presigned URL is reachable by an external client like the runner (matching the working web `ProjectAssetEndpoint`).

**Tests**
- Rust unit tests: MIME mapping, comment-HTML composition, image-component serialization, arg parsing.
- Python tests: endpoint selection for `is_server`/request, including a regression test for the `TypeError`.

## Scope note

The MCP comment tool is a hosted component that does **not** live in this repo, so the MCP half of the issue title is not addressed here. It would need the equivalent upload path added on the hosted MCP server.

## Validation

- `cargo test` (runner): 468 lib tests pass; clippy clean on touched files.
- `pytest pi_dash/tests/unit/settings/test_storage.py`: 12 pass.
- Manual end-to-end (upload against a live MinIO stack + render in the web UI) not run here to avoid writing to a live workspace — steps are in the issue's test hand-off comment.


irichard00:
## Review pass — approved, one fix applied

Reviewed the full diff against the source it depends on, and traced the embed path end-to-end in the frontend. The core design holds up:

- `GET /api/assets/v2/workspaces/{slug}/projects/{pid}/{asset_id}/` (`ProjectAssetEndpoint.get`) only requires `workspace__slug`, `project_id`, `pk` and `is_uploaded` — no comment binding, exactly as the PR assumes. The CLI sets `project_id` at create and `is_uploaded` at step 3, so the asset is servable.
- `getAssetSrc` in `apps/web/core/hooks/editor/use-editor-config.ts:73` passes a non-`http` `src` straight to `getEditorAssetSrc({assetId, projectId, workspaceSlug})`, so a bare asset UUID in `src` is the right thing to emit.
- The omitted `status` attribute defaults to `ECustomImageStatus.PENDING`, which is harmless: `node-view.tsx` only branches on `DUPLICATING`/`DUPLICATION_FAILED`, and `block.tsx:217` gates rendering on `resolvedImageSrc`, not on `status`.
- Every MIME the CLI emits (jpeg/png/gif/svg+xml/webp/tiff/bmp) is in `settings.ATTACHMENT_MIME_TYPES`, so there's no client/server allowlist mismatch.
- The `is_server` fix is real: `S3Storage.__init__(self, request=None)` on `main` genuinely has no such parameter, so `S3Storage(request=..., is_server=True)` raised `TypeError` on every call.
- `post_multipart` correctly sends no credentials — `ApiClient`'s `reqwest::Client` is built with no default headers and `X-Api-Key` is added per-request in `request()`. The presigned `fields` (which include `key` and `Content-Type`) are appended before the `file` part, as S3 requires.
- `PATCH .../assets/{id}/` returns 204 with an empty body; `request()` maps an empty success body to `Value::Null`, so step 3 doesn't trip.

### Fixed in this pass (pushed as a follow-up commit)

**Images were uploaded one at a time with no up-front validation.** `pidash comment add ISSUE --image good.png --image notes.txt` created the asset row for `good.png`, pushed its bytes to S3, and set `is_uploaded=true` — and only then hit the extension check on `notes.txt` and aborted. No comment was posted, so nothing ever referenced `good.png`; it stayed in the workspace bucket as an orphan. Same for any unreadable path later in the list.

Fixed by reading and validating every `--image` into a `PreparedImage` before the first upload, so a bad path or unsupported type anywhere in the batch fails before anything is written. Also reclassified an unreadable image file as `EXIT_INVALID` rather than `EXIT_UNKNOWN`, to match the unsupported-type case. Three unit tests added; runner lib suite is 471/471 green, clippy clean.

### Two things left for you to decide — neither blocking

**1. `UserServerAssetEndpoint` (`api/views/asset.py:336`) is the third `is_server=True` call site and was left as-is.** With the new semantics it now signs against the internal `AWS_S3_ENDPOINT_URL` (`http://pi-dash-minio:9000`) while returning `upload_data` to an external API-key client — a URL that caller cannot reach on a self-hosted MinIO deployment. It went from an obvious `TypeError` 500 to a 200 carrying an unusable URL, which is harder to diagnose. There's no in-repo caller (the web app uses the v2 `/api/assets/v2/user-assets/` route instead), and it's outside this PR's stated acceptance criteria, so I left it alone rather than widen the diff. If the intent is that it behaves like `GenericAssetEndpoint`, the same one-word change applies. (No effect on non-MinIO/AWS deployments — `is_server` is ignored on that branch.)

**2. Images over `FILE_SIZE_LIMIT` (5 MB) fail with a raw S3 XML error.** `GenericAssetEndpoint.post` does `size_limit = min(size, settings.FILE_SIZE_LIMIT)` and signs a policy with `content-length-range [1, size_limit]`, then stores that truncated value as the asset's `size`. Upload a 10 MB PNG and the server happily returns a presigned POST whose policy guarantees rejection; the CLI surfaces S3's `EntityTooLarge` XML. This is pre-existing server behavior, but this PR is what makes these endpoints reachable for the first time, so it's newly hittable. The clean fix is a 400 from the endpoint when `size > FILE_SIZE_LIMIT`; worth a separate issue rather than folding in here.

The live end-to-end upload still hasn't been run — that remains the one thing only a real stack can confirm.

PR #408, opened 2026-09-17
PDASHOSS01-172 Always-on built-in engine: codex Bridge multiplexer (Step 1, slice 1)

irichard00:
## What & why

First slice of PDASHOSS01-172 (always-on built-in engine). The built-in codex
engine is spawned per live session today — one `Bridge` = one subprocess = one
thread — so two chats mean two engines. This slice makes `codex::Bridge` a
**session → thread multiplexer** so one long-lived engine process can host many
conversations/runs as separate threads. It is the self-contained, no-protocol-change
core the issue calls out for Step 1 ("test this first, against the fake app-server harness").

## Changes

- **`fix(runner)`: add the missing `MuseCode` arm to `agent_kind()`** (first commit).
  The runner lib does **not compile at HEAD** — `AgentCursor::agent_kind()` omits its
  `MuseCode` arm and the variant isn't feature-gated, so the non-exhaustive match is a
  hard error and blocks every runner build/test. This commit is a standalone prerequisite
  hotfix and is safe to cherry-pick onto `main` on its own. **Heads-up: `main` is currently red.**
- **`feat(runner)`: `codex::Bridge` multiplexer core.** `warm_session` / `run_session` /
  `release_session` acquire, drive, and close per-session codex threads; the Bridge owns a
  `session→thread` map and a `thread→cursor` map; `next_session_event()` demuxes inbound
  frames by the `threadId` codex stamps, preserving per-thread ordering and preventing one
  busy turn from starving another. The legacy single-`thread_id` `warm`/`run`/`next_frame`
  path is untouched, so `agent::AgentBridge`, `ipc::chat`, and `daemon::supervisor` keep
  compiling and behaving exactly as before.

## Not in this slice (follow-ups, tracked on the issue)

- Migrate callers onto the multiplexer and make the daemon hold **one shared engine** across
  the chat lane + issue-run lane (today separate loop-local bridges in separate worker tasks).
- Crash/resume of live threads from stored ids; idle thread-close + RSS recycle threshold.
- **Step 2 (one daemon per app)** needs the desktop `managed_runner.rs` daemon-collapse **and**
  a cloud enrolment change (workspace-scoped machine tokens, PRIVATEPI1-60); it cannot fully
  land in this repo.

## Testing

- New `multiplexed_sessions_route_frames_to_their_own_threads` in the fake app-server harness:
  two threads stream interleaved, thread-stamped frames through one process; each cursor sees
  only its own events, in order; releasing one session leaves the other live.
- `cargo test --lib` (471 passed), `cargo test --test codex_bridge_fake` (8 passed), `cargo clippy --all-targets` clean.


irichard00:
## Review — PDASHOSS01-172 shared codex engine (Step 1)

Reviewed the Step-1 slice stack (`cce87e44..a184ae28`) against the acceptance criteria. Strong, well-sliced work; the engine actor, multiplexer, and fake-harness coverage are solid. One **confirmed correctness defect** gates the "crash → resume" criterion, plus a few non-gating notes.

### 🔴 Confirmed defect — chat crash/resume is defeated by an ungated teardown (blocking)

`runner/src/daemon/supervisor.rs:1982` (`handle_warm`) and `runner/src/daemon/supervisor.rs:2100` (`handle_turn`) both run:

```rust
if bridge.as_ref().is_some_and(bridge_has_exited) {
    if let Some(bridge) = bridge.take() {
        bridge.shutdown(Duration::from_secs(1)).await.ok();
    }
    *started_sent = false;
}
```

Slice 3 (commit `6d3e5c49`) added `survives_process_exit()` and gated **only** the in-loop crash branch (`supervisor.rs:2254`) so a mid-turn crash keeps the runtime warm instead of releasing the session. But these two *pre-operation* guards were not gated. After a shared-engine crash:

1. Keep-warm returns `Ok(false)`, so the `SharedCodex` bridge stays `Some` and the actor retains `resume_map[session]`.
2. The next message enters `handle_turn`/`handle_warm`. The process handle still points at the dead process (respawn is lazy; `obs.process_handle` is only swapped in `ensure_alive`, `engine.rs:747`), so `bridge_has_exited` is **true**.
3. `bridge.take().shutdown()` → `AgentBridge::SharedCodex` shutdown calls `handle.release_session(...)` (`agent/mod.rs:632`) → the actor's `Release` handler does `resume_map.remove(&session)` (`engine.rs:615`).
4. The bridge is rebuilt via `build_chat_bridge` with the same `session_key` but **no** resume id (the shared path intentionally ignores `--resume`, `supervisor.rs:1926`). `warm_session` respawns the engine, but with `resume_map[session]` gone it starts a **fresh thread** → the conversation is lost.

This is exactly the failure slice 3 was built to prevent ("killing the engine mid-session … resumes each live thread from its stored id, with the user seeing a failed turn rather than a lost session"). It isn't caught by the fake tests because they drive the engine actor directly, not the supervisor's `handle_turn`/`handle_warm` flow.

**Suggested fix:** gate both teardown blocks the same way the crash branch is gated — skip teardown for a self-healing shared engine and let the actor respawn + resume:

```rust
if bridge.as_ref().is_some_and(|b| bridge_has_exited(b) && !b.survives_process_exit()) {
    ...
}
```

This needs a **supervisor-seam regression test** (crash a shared engine mid-chat, send a follow-up, assert the same thread id resumes) so the integration path is actually covered — the current gap is precisely that this path has no test.

### Non-gating notes (for In Test / follow-ups, not blocking)

- **Process-wide interrupt (`agent/mod.rs:616`).** A cancel on the shared engine hits `turn/interrupt`, which is not thread-scoped, so one lane's cancel can abort the other lane's in-flight turn. Already documented and scoped to **PDASHOSS01-162**; event streaming stays isolated. Noting for completeness, not blocking this issue.
- **Resumed session not counted as live (`bridge.rs:184` vs `live_thread_count`, `bridge.rs:205`).** `resume_session` inserts into `sessions` but not `cursors`, so a just-resumed session reads as idle and is technically recycle-eligible. Practical impact is ~nil (resume runs on a fresh ~115 MB process, well below the 1 GiB recycle threshold), but the two bookkeeping maps drifting is worth a comment or a `sessions`-based `live_thread_count`.
- **Unstamped-frame fallback drops frames when >1 thread is live (`bridge.rs:224`).** With the shared engine, two live threads is the common case, so any routable frame codex emits *without* a `threadId` would be dropped rather than delivered. Fine under the design premise that all routable frames are stamped — worth confirming against the real bundled app-server in In Test.
- **Respawn cap reset by any progress (`engine.rs:572`).** An engine that completes the handshake but always crashes mid-stream resets `respawns_since_progress` each cycle, so it never trips `MAX_RESPAWNS_WITHOUT_PROGRESS`. Respawn is user-driven and backoff-throttled, so this is low severity / arguably by design, but the module doc's "dies again before any frame flows" wording overstates the guarantee.

Only the first item blocks. Sending back to In Progress to land the fix + a regression test.

PR #409, opened 2026-09-17
PDASHOSS01-173 Approvals for the built-in agent: surface the engine's prompts in chat, and let the user choose the mode

irichard00:
## Why

The built-in codex engine started every thread with a hardcoded `sandbox=danger-full-access` / `approval_policy=never` (`codex/bridge.rs`), overriding the managed config and meaning **nothing ever asked the user** before running shell commands. The whole approval path already wired through `chat.rs`, `pidash-ipc` and the desktop transport was dead code because the engine never asked.

This builds the approval system so a safer default becomes possible. **Full access stays the default** — this is not flipping the switch.

Answers PDASHOSS01-162: the per-thread `thread/start` sandbox/approval value **does** override the managed `config.toml`, so today's behaviour was full access with no prompts and no denylist enforcement.

## What changed

**Runner**
- New `ApprovalMode` (`ask` / `workspace` / `full_access`) + `engine_thread_settings` mapping (`runner/src/approval/mode.rs`). The codex `Bridge` stamps the derived `sandbox`/`approval_policy` on `thread/start`; its default stays the legacy `(danger-full-access, never)` so the **cloud path is unchanged**.
- `Policy` is now mode-aware: the denylist refuses in **every** mode (hard floor), full access auto-approves the rest (no user prompts), ask/workspace surface the request. Every mode uses `approval_policy=on-request` so the runner stays in the loop to enforce the denylist — that is how "no prompts under full access" and "denylist refused regardless of mode" both hold.
- The mode is captured when a session's bridge is spawned and held for the thread's life, so changing it never retunes a turn already in flight.

**IPC / desktop**
- `mode` rides on `Request::ChatWarm` / `ChatSend` and the `chat_warm` / `chat_send` Tauri commands (absent = legacy posture, back-compatible).

**Web**
- `RunnerChatPage` renders an inline approval prompt (Approve / Deny / "Always allow this session") parked with the request's TTL and cleared on decision, terminal frame, or interrupt.
- A mode control in the header, persisted per runner, gated to the local transport. Cloud chat and the `/runners/approvals` page are untouched.
- Wording lives in the `apps/web/ce/components/desktop/` seam so editions can override it.

## Tests
- `runner/src/approval/mode.rs` + `policy.rs` unit tests, incl. `denylist_refused_in_every_mode` and the ask/workspace/full-access mapping.
- Desktop transport tests assert `chat_warm`/`chat_send` carry the chosen mode and default to `full_access`.
- `cargo test -p pidash -p pidash-ipc` (approval/chat/codex-bridge), web typecheck + lint, desktop overlay vitest (48), and `desktop/src-tauri` `cargo check` all green.

Related: PDASHOSS01-159, PDASHOSS01-162, PDASHOSS01-172, PDASHOSS01-107.


irichard00:
## Code review (PDASHOSS01-173 — In Review pass)

Reviewed all 19 files against the issue scope and acceptance criteria. **No blocking defects found — approving to move on to In Test.** CI is green (5/5 required checks).

### What I verified
- **Mode → engine mapping (`runner/src/approval/mode.rs`)** — `ask`→`read-only`, `workspace`→`workspace-write`, `full_access`→`danger-full-access`; all three use `approval_policy = on-request`. The reasoning (runner must stay in the loop so the denylist can fire even under full access, while still showing no prompts) is sound and documented. Default = `LEGACY_FULL_ACCESS` (`danger-full-access`/`never`) for the mode-less path.
- **Cloud path unchanged** — `ChatWorker`/`AssignWorker` spawn with `None` mode → legacy posture (engine `never`, no requests surfaced). Their `Policy::new(..., ApprovalMode::Ask)` reproduces the previous mode-less default exactly: I diffed the old `evaluate` and confirmed the command allow/deny/readonly checks and the `FileChange`/`NetworkAccess` fallbacks all resolve to the same `Decision` as before for `Ask`. Behaviour-preserving.
- **Denylist floor in every mode** — `evaluate` refuses denylisted commands before the mode default, in all three modes (`denylist_refused_in_every_mode` covers it). `AutoDecline` → `into_cloud()` → engine decline, no prompt.
- **Mode captured at spawn, held for thread life** — `SessionRuntime.approval_mode` is set in `ensure_bridge` only when the bridge is first spawned; `resolve_approval` reads the captured value via `DriveCtx`/`ApprovalCtx`, so a mid-session mode change can't retune a turn in flight. Matches "affects the next thread".
- **Wire/transport plumbing** — `mode` rides `ChatWarm`/`ChatSend` with `#[serde(default, skip_serializing_if)]` (back-compat verified by the omit test); TS `TApprovalMode` matches the serde snake_case. Transport `chat_approval_request` frame keys (`local_approval_id`, `approval_kind`, `reason`, `expires_at`, spread payload) line up exactly with `pendingApprovalFromEvent`'s destructure and `approvalTarget`'s payload reads.
- **UI: park/decide/TTL/cancel/gate** — inline prompt cleared on decision (optimistic, restored on error), on any terminal frame, on the client-side TTL timer, and on `stop()`; the whole approvals surface is gated on `decideChatApproval` so the cloud transport and `/runners/approvals` are untouched.

### Non-blocking observations (no action required for this issue)
- Changing the mode mid-session updates the transport slot but has no visible effect until a *new* chat/thread spawns (the runner holds the captured mode for the bridge's life). This is exactly the documented "affects the next thread" semantic, but a user changing the control during a warmed session sees no immediate change — worth a tooltip/disabled-state hint in a future polish pass.

The one thing this pass can't cover is the live desktop click-through (steps 1–7 of the test plan) and the 10-minute wall-clock TTL — those belong to In Test.

PR #416, opened 2026-09-17
PDASHOSS01-176 Peek overview "expand" button does not open the work item detail page

irichard00:
## Summary

The peek-overview **expand** control (diagonal-arrow icon, "Open work item in full screen") did nothing useful — clicking it left the user on the issues list instead of navigating to the work item detail page.

### Root cause

`apps/web/core/components/issues/peek-overview/header.tsx` rendered the control as `<Link href={workItemLink} onClick={() => removeRoutePeekId()}>`. In this app `next/link` is a compat shim (`app/compat/next/link.tsx`) over react-router, and `peek-overview/url-sync.tsx` drives the URL with react-router `setSearchParams(..., { replace: true })`. Clicking cleared the peek store **synchronously**, and the resulting store→URL `replace` raced the Link's `push`, so the pushed `/browse/...` entry was clobbered / the list location re-asserted.

### Fix

Decouple teardown from navigation. On a plain primary click: `preventDefault`, `removeRoutePeekId()`, then `router.push(workItemLink)`. The compat `useRouter().push` defers navigation by a macrotask (`setTimeout(0)`), so the param-stripping `replace` (a React passive effect) lands **first**; the push then wins and the list history entry is left without `peekIssueId`. Modified clicks (meta/ctrl/shift/alt/non-primary) fall through to the native `<a href>` so open-in-new-tab still works and does not tear down the peek.

Also guard against navigating to `/browse/undefined-undefined/` while the issue detail is still loading — the control is disabled until the link resolves (archived items route by projectId/issueId; regular items need the loaded `projectIdentifier` + `sequence_id`).

## Changes

- `apps/web/core/components/issues/peek-overview/header.tsx` — programmatic clear+push, native modified-click passthrough, malformed-link guard.
- `apps/web/tests/issues/peek-overview-header.test.tsx` — new component test (navigation, order, archived path, modified-click, disabled-while-loading).
- `apps/web/vitest.config.ts` — mirror the `next/*` compat aliases from `vite.config.ts` so component tests can resolve the shims.

## Validation

- `pnpm turbo run check:types --filter=web` — pass
- `pnpm turbo run check:lint --filter=web` — 0 errors
- `pnpm vitest run` — 22 files, 117 tests pass (4 new)


irichard00:
## Review — PDASHOSS01-176 peek expand button

Reviewed the fix against the issue's acceptance criteria. The core race fix is sound:

- **Race fix (confirmed correct):** `handleOpenFullScreen` `preventDefault`s the plain primary click, calls `removeRoutePeekId()`, then `router.push(workItemLink)`. `useAppRouter().push` (→ `next/navigation` compat shim) defers via `setTimeout(0)`, so the store→URL sync's param-stripping `replace` (a React passive effect) lands before the push. This clears `?peekIssueId` from the list history entry and leaves the pushed `/browse/...` entry intact — satisfying both the navigation and the clean-Back criteria.
- **Modified clicks (confirmed correct):** meta/ctrl/shift/alt/non-primary clicks return early without `preventDefault`, falling through to the native `<a href>` for open-in-new-tab, and never tear down the peek.

### One defect found — and fixed in this branch (25d64ae9)

The malformed-link guard was inconsistent with the link it guards, for **archived** items:

- `isWorkItemLinkReady` (archived branch) trusts the route **prop** `projectId` — present immediately.
- but `workItemLink` was built from `issueDetails?.project_id` — `undefined` until the detail fetch resolves.

So while an archived issue's detail was still loading, the expand button was already enabled and its href resolved to `/<ws>/projects/undefined/archives/issues/<id>` — a malformed URL, which is exactly what acceptance criterion 6 rules out (the browse path was covered; the archive path was not). The existing archived test only exercised the already-loaded case, so it passed.

**Fix applied:** feed `projectId ?? issueDetails?.project_id` to `generateWorkItemLink` (the route projectId is the peeked issue's project and is what the guard already trusts), and added a test covering archived-while-loading. `pnpm vitest run` on the file → 5 pass; `check:types` pass; `check:lint` 0 errors.

Remaining acceptance criteria (Back button, new-tab, and the cycle/module/workspace/archived variants in a real browser) are unit-reasoned but not driven end-to-end here — they carry into the In Test stage per the hand-off comment.

irichard00:
## Second review pass — PDASHOSS01-176 (HEAD `46cdbf15`)

Re-reviewed after `46cdbf15` landed (the first pass only covered through `25d64ae9`). **Approving — no unresolved findings.**

### The url-sync fix is correct, and load-bearing

`url-sync.tsx:54` — `searchParams.get(PEEK_NESTING_QUERY_KEY) || undefined`.

The diagnosis in the commit message holds up against the code: `urlPeekNestingRaw` feeds the store→URL effect's "nothing to change" guard at line 109 (`desiredNestingParam === urlPeekNestingRaw`), and `desiredNestingParam` is `string | undefined` (line 104). With the raw value left as `null`, that comparison was `undefined === null` → `false` on *every* run for any peek without a nesting level — i.e. the overwhelmingly common case — so the effect issued a `replace` unconditionally. The other two URL params on lines 46–47 were already normalized the same way; this was the odd one out.

I checked the guard for behaviour regressions in both directions and it stays correct:
- URL `peekNestingLevel=2` → store nesting cleared: `undefined !== "2"` → still writes (param deleted). ✅
- No URL param → store nesting 2: `"2" !== undefined` → still writes (param set). ✅
- `urlPeekNestingLevel` on line 55 was already a truthiness check, so `null` → `undefined` doesn't change it. ✅

**Verified the tests are real, not tautological:** I reverted the one-liner locally and re-ran `tests/issues/peek-overview-url-sync.test.tsx` — exactly the 3 idempotence cases fail (`3 failed | 3 passed`), the 2 "still writes" cases and the nesting case keep passing. Restored afterwards.

### Independently re-ran the browser walkthrough

Not taking the 14/14 on trust — I checked the branch out, confirmed the dev server on :3000 was serving it, and ran `.pidash/e2e/verify.mjs` myself against the local stack: **14 pass / 0 fail**, covering C1 (project list → `/ai-republic/browse/AIREP-2/`), C2 + no-peek-reopen, C3 + peek-kept, C4 cycle/module/workspace-all-issues, C5 archived → `/…/archives/issues/<id>/` + back, C6, and the close/reopen/Properties smoke.

### header.tsx re-check (previously reviewed, no new issues)

- `projectId ?? issueDetails?.project_id` — `projectId` comes from `peekIssue.projectId` via `view.tsx:177`, so it's the peeked issue's own project. Correct source, and it resolves the archive-link-while-loading defect from the first pass.
- Modified-click passthrough: `next/link` compat applies `ensureTrailingSlash(href)` itself (`app/compat/next/link.tsx:20`), so the rendered anchor for an archived item is `/…/archives/issues/<id>/` — the same URL `router.push` produces. Cmd-click and plain click can't diverge on the trailing slash. (`generateWorkItemLink` emits the archive path without one.)
- `isWorkItemLinkReady` is narrower than it looks: `view.tsx:161` only mounts the header once `getIssueById(issueId)` resolves, so the guard's live window is really "project store hasn't produced `projectIdentifier` yet" (workspace-level views). Still worth having; just noting the C6 browser check can't exercise the disabled branch, which matches the harness output (`control=A aria-disabled=null`).

### One non-blocking note

`|| undefined` also folds an empty `?peekNestingLevel=` to `undefined`, so a hand-crafted URL with that empty param would no longer be stripped by the sync. Cosmetic and unreachable through the UI — flagging it only so it isn't a surprise later.

### Gates re-run at `46cdbf15`

`pnpm --filter web exec vitest run` → 23 files / **124 pass** · `check:types` clean · `check:lint` **0 errors** (881 warnings, under the gate) · `git diff origin/main...HEAD` is 5 files, scoped to the peek overview + its tests + the vitest `next/*` aliases.

Moving PDASHOSS01-176 to In Test.

PR #422, opened 2026-09-18
PDASHOSS01-177 Muse Code runner fails 100% of runs: the bridge parses an invented JSONL schema

irichard00:
Fixes PDASHOSS01-177: every Muse Code run was reported as failed even when the agent finished its work.

## Root cause
`runner/src/muse_code/schema.rs` dispatched on a top-level `type` field. The real `muse exec --json` output has no such field: every line is a `record_type` / `payload_type` / `payload` envelope. So every frame was classified `unknown` and `run.terminal.completed` was never seen. The clean exit then reached the watchdog as "agent stdout closed". The old fake test used the same made-up schema, so the suite stayed green.

## Changes
- **schema.rs** now models the real envelope. It was built from frames captured from `muse 1.3.0-R3401.1`.
- **Terminal records:**
  - `run.terminal.completed` becomes `Completed`, carrying `text` as `result`.
  - Any other `run.terminal.*` becomes `Failed`, with the reason in its detail.
  - `task.lifecycle.failed` is a task-level event (for example, one failed bash call) and does not fail the run.
- **Transcript events are now typed** and reuse the method names the daemon already understands:
  - `run.output.delta` fragments are joined into one `assistant/message`. It is picked up by `extract_agent_message_text` and reaches the cloud as `agent/message`.
  - `tool.result` becomes `user/toolResult`, with the call id, tool name, outcome and `is_error`.
  - A `task.lifecycle.proposed` for a `tool.*` task becomes `tool_call/started`.
  - Other records pass through as `muse/<payload_type>`. `run_event_mirror` compacts them, because there are about 500 per run.
- **`wait_for_init` is strict.** It buffers records until `run.lifecycle.started` (or an early `run.terminal.*`). It **bails on the first frame that isn't a record envelope**, with a message that names the schema, and bails on EOF. The accept-any-first-frame fallback is gone.
- **Session id:**
  - It now comes from the session stream UUID (`stream.kind == "session"`). The `muse-{run_id}` synthesis is removed, and the fallback is the bare run id.
  - `build_command` also drops any non-UUID session id, such as the legacy `muse-…` pointers stored by older runners, instead of letting `muse exec` exit 2.
- **Tests** now replay fixtures captured from the real binary (`runner/tests/fixtures/muse_code/`):
  - a Meta-provider run, with prompt, tool output and answer text redacted;
  - two `--provider echo` turns, the second resumed with `--session-id`;
  - a real `run.terminal.failed`.

## Validation
- `cargo test` (runner): all suites pass, 475 lib tests. That includes the 12 muse lib tests and the 10 tests in `muse_code_bridge_fake`.
- Live check, done locally and not committed: the real `muse exec --json --provider echo` (offline, free) was driven through `Bridge::run_with_command` for two turns.
  - Turn 1 completed and captured the session UUID.
  - Turn 2 was resumed with `--session-id <that UUID>`, kept the same thread id and completed.
  - No `unknown` events appeared.
- Reproduced by hand: `muse exec --session-id muse-<uuid>` exits 2 with "expected a UUID".

## cursor_agent audit
`cursor_agent` doesn't have Muse's blind spot. Its `wait_for_init` waits for `system/init` (or a pre-init `result`), buffers everything else and fails loudly on timeout or EOF, and its stream-json schema is Cursor's documented format. One latent issue, left for a follow-up: after a failure before init it synthesizes `cursor-{run_id}`, which would then be passed to `--resume`.


irichard00:
## Review: PDASHOSS01-177 (Muse record envelope)

I reviewed `e7062258` against the issue's acceptance criteria and found no defects. Approving.

What I checked:
- **Schema** (`muse_code/schema.rs`): the parser dispatches on `record_type`/`payload_type`, and a frame that isn't a record envelope becomes `Unknown`. `session_id()` only accepts a `stream.kind == "session"` UUID.
- **Terminal mapping**: `run.terminal.completed` becomes `Completed`, with `result` set to the terminal text. Any other `run.terminal.*` becomes `Failed`, with Muse's `reason` in the detail. `task.lifecycle.failed` passes through as `muse/task.lifecycle.failed` and does not end the run.
- **Transcript**: the `run.output.delta` fragments are joined into one `assistant/message` (the text-block shape the daemon's `extract_agent_message_text` already reads). `tool.result` becomes `user/toolResult` with `tool_use_id` and `is_error`, which the exec-completion hints in `observability.rs` consume. A `task.lifecycle.proposed` whose kind is `tool.*` becomes `tool_call/started`; I confirmed `task_id` sits at the top level of the payload in the real fixture. The `muse/*` bookkeeping records are compacted in `run_event_mirror`.
- **Strict `wait_for_init`**: it bails on a non-envelope first frame, on EOF, and on a 30s timeout. It returns on `run.lifecycle.started`, or on an early `run.terminal.*` so the real failure reason still gets through.
- **Session id**: it comes from the session stream UUID and falls back to the bare `run_id`. `build_command` drops a legacy non-UUID id instead of passing it to `--session-id`.
- **Fixtures**: I scanned them for local paths, usernames, emails and key-like strings and found none.
- **Locally**: `cargo test --lib muse_code` (12) and `cargo test --test muse_code_bridge_fake` (10) pass, and `cargo clippy --all-targets -D warnings` is clean. The CI `cargo test + clippy` job was still running when I reviewed.

Non-blocking follow-ups (already noted in the PR): `cursor_agent`'s synthesized `cursor-{run_id}` could reach `--resume` after a failure before init; and the daemon's endless retry on a 401 `runner_not_found` is out of scope here.

PR #423, opened 2026-09-19
PDASHOSS01-148 Voice dictation for agent input (BYO speech-to-text endpoint)

irichard00:
Fixes the blocking defect found while testing the voice dictation epic (PDASHOSS01-148): push-to-talk never transcribed from the UI.

**Root cause.** `AssistantService.transcribeAudio()` uploaded the recording under the multipart field `audio`, but `assistant/views/transcribe.py` reads `request.FILES.get("file")`, which is the OpenAI-compatible field name. Every hold-and-release returned `400 no_audio` on all three chat surfaces. The existing tests missed it because the hook test mocks `transcribeAudio` and the API contract tests post `file`.

**Changes**
- `packages/services` — `transcribeAudio` now sends `file`. The docstring now names the real error body (`{error: "stt_config_missing"}`).
- `use-dictation.tsx` — the transcribe-error fallback now matches `error === "stt_config_missing"`, the code the API actually returns. It was checking `error_code === "not_configured"`, which the API never sends, so users got a generic error instead of the "Set up voice dictation" state.
- `dictation-settings.tsx` — when the connection test fails, the toast shows the readable `detail` and falls back to `error_code` only when there is no detail.
- `transcribe.py` — docstring only. The real upload cap is `FILE_SIZE_LIMIT` (5 MB by default), enforced by `RequestBodySizeLimitMiddleware` before the view's 25 MB check runs. That is ample for the recorder's 60 s opus cap.
- Tests:
  - A new service-level test checks the multipart field name. It fails on the old `audio` field.
  - A hook test covers the `stt_config_missing` fallback.
  - A settings test checks that the toast prefers `detail`.

**Validation**
- `cd apps/web && npx vitest run tests/chat tests/assistant`: 34/34 pass.
- tsc: no errors in the touched files.
- oxfmt and oxlint: clean.


irichard00:
Review (PDASHOSS01-148): no findings, approved.

- `transcribeAudio` now appends `file`, which is what `transcribe.py` reads (`request.FILES.get("file")`). The new service test pins the field name and checks that `audio` is absent.
- Hook fallback: the service rethrows `err.response.data`, and the view returns `{"error": "stt_config_missing", ...}`. The hook now reads `.error`, so the two match.
- Settings toast: `stt-config/test/` returns `{ok:false, error_code, detail}` when the provider rejects the request. It returns `error_code` alone on the early-exit paths, so `detail || error_code` covers both.
- The only API change is a docstring.

CI: lint, web and desktop tests pass. `pytest` was still running when I checked; the only backend change here is a docstring.

PR #425, opened 2026-09-20
PDASHOSS01-185 Let agents read project pages: /api/v1/ page read endpoints and `pidash page` CLI commands

irichard00:
Closes PDASHOSS01-185.

Project pages work as a project-scoped wiki for people, but every route an agent uses to reach Pi Dash was blind to them. `/api/v1/` had no page routes (pages were served only by the first-party session API in `apps/api/pi_dash/app/urls/page.py`), so neither the `pidash` CLI nor the MCP connector — which wraps `/api/v1/` — could see one. This is the read path. It is read-only; creating, editing, locking and archiving stay out of scope, as do page search, versions, workspace-level (`is_global`) pages, and pinning a page into a project's prompts.

## 1. Read endpoints on `/api/v1/`

```
GET /api/v1/workspaces/{slug}/projects/{project_id}/pages/
GET /api/v1/workspaces/{slug}/projects/{project_id}/pages/{page_id}/
```

Both follow the conventions of the existing v1 views — `X-Api-Key` auth, per-token throttling, cursor pagination, `fields`/`expand`, drf-spectacular schema. The list returns metadata only (`id`, `name`, `parent`, `owned_by`, `access`, `is_locked`, `archived_at`, `created_at`, `updated_at`); the detail adds `description_html`, `description_stripped` and a new `description_markdown`.

**Access rules mirror `apps/api/pi_dash/app/views/page/base.py`.** The caller must be an active member of the project and the project must not be archived; a page is visible when it is public or the caller owns it.

Two decisions worth review:

- **Someone else's private page is a 404, not a 403.** The visibility rule lives in the queryset (`Q(access=0) | Q(owned_by=user)`) rather than in a permission class, so the page simply isn't there and `.get()` 404s. `ProjectEntityPermission` supplies the 403-for-non-member half. Note this differs from the first-party `ProjectPagePermission`, which reaches for the page before the owner check and would answer 403 — the ticket asks for 404, and 404 is also the answer that doesn't confirm the page exists.
- **Project scoping uses `Exists(ProjectPage…)`, not a join through the `projects` m2m.** An m2m join fans the page row out once per linked project, which would let "linked to this project", "project not archived" and "caller is a member" each match on a *different* project.

**On `page_view`:** the ticket asks what the first-party view does with a project whose `page_view` feature is off. The answer is *nothing* — `page_view` appears only in the project serializer's and project view's field lists; no page read path consults it. These endpoints do the same, and the view docstring records that so the next reader doesn't have to re-grep.

The include-archived flag is spelled `include_archived` (matching the CLI's `--include-archived`); `format` is reserved by DRF for renderer selection, per the ticket.

## 2. HTML → markdown

Pages are stored as Tiptap HTML, which agents read poorly. `pi_dash.utils.markdown_converter.html_to_markdown` is the one server-side renderer behind the API, the CLI and the future MCP tool, so they cannot disagree. It is pure — no DB, no request state — and importable from anywhere.

**Library: [`markdownify`](https://pypi.org/project/markdownify/) 1.1.0.** MIT-licensed, pure Python, and built on `beautifulsoup4`, which is already a dependency — so it adds one small wheel and no second HTML parser. It is also a plain subclassable converter (one `convert_<tag>` method per tag), which is what lets us teach it the Tiptap-only nodes without forking it. (`html2text`, the other obvious candidate, is GPLv3 and hand-rolls its own parser.)

`TiptapMarkdownConverter` adds:

- task lists — Tiptap spells them as `<ul data-type="taskList">` / `<li data-type="taskItem" data-checked>`, not a distinct tag;
- `mention-component` → `@user_mention:<id>` (the display name would need a DB lookup this helper deliberately does not do);
- `image-component` → ``, falling back to `[image <id>]`;
- `issue-embed-component` → a relative work-item link, falling back to `[work item <id>]`;
- fence languages read off `<code class="language-…">`.

Unknown nodes keep their text rather than vanishing, and the helper never raises — a conversion failure degrades to the tag-stripped text so one odd document can't fail a read. Escaping of `_` and `*` is off: this markdown is read, not round-tripped, and escaping mangles identifiers like `snake_case_name`.

## 3. `pidash page` CLI

```
pidash page list --project <PROJ-or-UUID> [--cursor <c>] [--per-page <N>] [--include-archived]
pidash page get <page-uuid> --project <PROJ-or-UUID> [--body-only]
```

Same contract as `issue`/`workpad`: JSON on stdout, JSON errors on stderr, `api_client::EXIT_*` codes. `--body-only` prints just the markdown so an agent can redirect it into a file. A non-UUID page id is rejected locally rather than 404-ing on the route's `<uuid:page_id>` converter. `build_query_string` in `cli/issue.rs` became `pub(super)` so both list commands percent-encode identically; `cli::page` is `pub mod` so the existing HTTP-fake contract test can drive `cmd_list`/`cmd_get`, matching how `cli::run_cmd` is already tested.

## 4. Telling agents

`prompting/sections/pidash-cli.md` and the legacy `prompting/fragments/03_pidash_cli.md` both gain a `#### Pages` block, and `docs/wiki/17-cli-reference.md` gains a full `## Pages` section.

**What an operator has to run: for the section registry, nothing.** Section defaults live in code and the DB stores only *overrides* (`prompting/registry.py`), so there is no seed or sync step — the next composed prompt picks the new text up. `pidash-cli` is `customizable: locked`, so no workspace override can shadow it. The `fragments/` copy feeds the *legacy* global default `PromptTemplate`; a deployment still relying on that row needs `python manage.py reseed_default_template --force`.

## Validation

- `pytest pi_dash/tests` — **1887 passed, 1 failed**. The one failure is `test_valid_agents_matches_runner_schema`, which fails identically on `main` (baseline: 1887→1844 passed, same 1 failure), verified by running both in the same container image off clean worktrees.
- 19 new contract tests (`tests/contract/api/test_pages.py`) and 24 new unit tests (`tests/unit/utils/test_markdown_converter.py`).
- `cargo test` — 488 lib + all integration suites pass, including 13 new `cli::page` unit tests and 6 new CLI contract tests.
- `cargo clippy --all-targets -- -D warnings` — clean.
- `manage.py spectacular` — `list_pages` and `retrieve_page` are in the schema with `description_markdown` and `include_archived`; zero page-related schema errors.
- `ruff check` / `ruff format` clean on every touched Python file.
- All seven new tracked files carry the `COPYRIGHT.txt` header.

The serializer and the markdown helper are both free of request-specific state, so the downstream MCP page tool can import them as-is.


irichard00:
## Review pass — approved, one fix applied

I read the whole diff against the ticket's acceptance criteria, traced the access rules through `ProjectEntityPermission` / `BaseAPIView.handle_exception` / the `Exists(ProjectPage…)` queryset, exercised the markdown converter against Tiptap HTML the unit tests don't cover, and regenerated the OpenAPI schema. One real finding, fixed in 6dc5d6e8; everything else holds up.

### Finding (fixed in 6dc5d6e8)

**The page list's OpenAPI example showed a generic placeholder, not a page.**
`create_paginated_response(PageLiteSerializer, "PaginatedPageResponse", …)` resolves its example through `get_sample_for_schema`, which strips `Paginated`/`Response` and looks `"Page"` up in `SCHEMA_EXAMPLES` (`utils/openapi/examples.py:889`). There was no `"Page"` key, so the published schema documented a page row as `{"id", "name": "Sample Item", "created_at"}` — every sibling v1 list endpoint documents its own shape. Separately, `PAGE_EXAMPLE` was defined and exported but referenced by no view, so it was dead.

Fix: lift the row into `SAMPLE_PAGE`, register it in `SCHEMA_EXAMPLES`, and build both `PAGE_EXAMPLE` and `PAGE_DETAIL_EXAMPLE` on it so the three can't drift. Verified by regenerating the schema — `list_pages`'s 200 example now carries the real nine-field page row.

### Verified

- **Access rules.** Non-member → 403 via `ProjectEntityPermission` (SAFE_METHODS branch keys off `view.project_id`, which `_rewrite_project_kwarg` has already normalised from a slug). Another member's private page is filtered out of the queryset, and `.get(pk=page_id)` raising `ObjectDoesNotExist` lands on `BaseAPIView.handle_exception` → 404. Archived project and cross-project pages both fall out of the `Exists` subquery.
- **The `Exists(ProjectPage…)` choice is right**, and worth keeping the comment on. The first-party view's `projects__project_projectmember__member=…` m2m join fans a page out per linked project; constraining membership, `archived_at` and `project_id` inside one subquery is what stops those three matching on three different projects. It also removes the `.distinct()` the first-party queryset needs.
- **Role parity.** `ProjectPagePermission._check_project_action_access` allows GET for ADMIN/MEMBER/GUEST — i.e. every active member — so `ProjectEntityPermission`'s active-membership check is not a widening.
- **Soft deletes** are excluded for free: `objects` is the `SoftDeletionManager`.
- **Custom node coverage is complete.** The editor registers exactly three custom tags (`mention-component`, `image-component`, `issue-embed-component` — `grep renderHTML packages/editor/src/core/extensions`), and the attribute names the converter reads (`entity_identifier`, `entity_name`, `src`) match the extension configs.
- **Markdown converter, beyond the unit tests.** I ran it over nested task lists, a task list nested in an ordered list, `<thead>` tables, Tiptap `colspan`/`colwidth` cells, mentions inside headings, images with `width`/`height`, an `<a>` with no `href`, entity-escaped code, `<br>`, and adjacent mentions. Nothing raised, nothing was dropped, and the checkbox never landed in the wrong place.
- **Exit codes** in `docs/wiki/17-cli-reference.md` match `api_client.rs`: 403 → 3, 404 → 4.
- **Prompt section.** `{% endif %}#### Pages` puts Pages outside the workpad conditional, same as `#### States` — so scheduler runs get it too, which is what you want.
- Tests green on the fix commit: 19 contract + 24 unit, in a fresh container off this worktree.

### Two notes, no action needed

- **Nested fences.** A page whose code block itself contains ```` ``` ```` renders as a fence inside a fence. That's markdownify's behaviour, it doesn't raise, and the text survives — flagging it only so it isn't a surprise later.
- **The two judgement calls in the PR body both look right to me.** 404 (not 403) for another member's private page is what the ticket asks for, and putting the rule in the queryset rather than a permission class is why it comes out that way. `include_archived` is a sensible name given `format` is off-limits and it mirrors the `--include-archived` flag.

PR #430, opened 2026-09-21
PDASHOSS01-192 Desktop: daily update check with a sidebar update button

irichard00:
## Summary
Pi Dash issue: [PDASHOSS01-192](https://pidash.airepublic.com/ai-republic/browse/PDASHOSS01-192)

The downloaded desktop app only checked for updates at launch, so anyone who kept it open for days never heard about a new version.

- **Daily check:** while the app runs, check once a day just after 00:00 UTC. Each app picks a random time in the first 15 minutes so they don't all hit `/api/desktop/updates/` at once. The loop re-reads the wall clock at least every 10 minutes. The tokio timer pauses while macOS or Linux is suspended, so without this a laptop could wake hours after midnight and still not check; it now checks once, soon after waking.
- **Sidebar update button:** an update found by the daily check is not prompted for. It shows a small corner icon to the right of the user menu at the bottom of the left sidebar. Clicking it downloads and installs the update, then restarts. A failure shows a toast and keeps the button so the user can retry.
- **Launch check unchanged:** same Install/Later dialog. "Later" now keeps the update so the sidebar button shows it.

### How it's wired
- `desktop/src-tauri/src/updates.rs` (new): the launch prompt (moved from `main.rs`), the daily loop, the `updater://available` event, and the `desktop_pending_update` / `desktop_install_update` commands.
- `apps/web/core/components/desktop-update-button.tsx`: the web build renders nothing here. `desktop-overlay/…/desktop-update-button.tsx` is the desktop version. It listens for the event and also asks for a waiting update on mount, because the sidebar remounts.
- `sidebar-wrapper.tsx`: the user menu row gives up a corner for the button.

Store (MSIX) builds are unaffected: they're built without the updater config, so neither check runs.

### Not in scope
- No button when the sidebar is collapsed (user menu moves to the top nav).
- Minimum supported version, and release notes in the prompt.

## Test plan
- [x] `cargo test`: 57 passed, including new tests for when the daily check is due (next UTC midnight, the day after, waking days later)
- [x] `desktop/scripts/test-overlay.sh`: 69 passed, including 5 new tests for the button (waiting update, event, install failure, unlisten)
- [x] oxfmt / oxlint clean on changed files
- [ ] Manual check on a signed build against a newer release: dialog at launch, button after "Later", click installs and restarts

`cargo clippy -D warnings` already fails on `main` (in `pidash_cli.rs` and the navigation handler in `main.rs`). This PR adds no new warnings.

https://claude.ai/code/session_01NKLm1U3bSi37gz7MTGXLoV


irichard00:
## Review — PDASHOSS01-192

Read the whole diff against the issue's scope. The design holds up: the daily
loop's short naps are the right answer to a monotonic timer that stops during
suspend, `desktop_pending_update` covers the event firing before the webview
listens, and `installing` correctly keeps a daily check from swapping `pending`
out from under a running install. Two things were wrong; both are fixed in
`ea22376`.

### 1. The install state didn't survive a sidebar remount

The component kept `installing` only in React state, but the code already knows
the sidebar remounts — that's why `loadPending` exists. Two ways to trigger it:

- collapse and re-expand the sidebar (`sidebar-wrapper.tsx:108` only renders the
  user-menu row when `!sidebarCollapsed`);
- navigate to `/:workspaceSlug/notifications`, which unmounts the sidebar
  outright (`workspace-app-sidebar.tsx:35` — `if (isNotificationsPath) return null`).

Downloading a full app bundle takes long enough to do either. On remount,
`loadPending` restored `update` but not `installing`, so the button came back
enabled. Clicking it hit the backend guard and surfaced
`Update failed — an update is already installing`, which reads as a broken
update rather than a double click.

Fixed by making the desktop the source of truth for both facts:
`PendingUpdate` gains an `installing` flag, `desktop_pending_update` reports it,
and the button applies it on mount and on every event.

That alone would have traded one bug for a worse one — a button restored to
`installing: true` would stay disabled forever if the download then failed,
because only the component that started the install sees the rejected promise.
So `desktop_install_update` now re-announces the update on failure, and any
mounted button drops back out of its installing state. Two new tests cover both
halves.

### 2. `subscribe()` had no rejection handler

`void subscribe()` versus `loadPending().catch(...)` right beneath it. A failing
`listen` became an unhandled rejection and silently left the component without a
listener. Now `.catch(() => undefined)`, matching its sibling.

### Verified

- `cargo test` — 56 passed, 1 ignored
- `desktop/scripts/test-overlay.sh` — 71 passed (was 69; +2 for the remount cases)
- `cargo fmt --check` and `cargo clippy` report nothing in `updates.rs`; oxfmt and
  oxlint clean on both changed `.tsx` files

The unchecked box in the PR description — the manual pass on a signed build
against a real newer release — is the one thing I can't do from here, and it's
the only way to exercise the actual download/install/restart path.

PR #431, opened 2026-09-21
docs(design): sub-identities — formal actor = (account, sub-identity)

irichard00:
## Summary

Design doc only — no code. Adds `.ai_design/create_sub_identity/design.md`.

Today every action in Pi Dash is recorded against a user account, so a comment written by a runner shows as "Rich M commented …" with a bold `Codex:` prefix in the body, and an issue created or edited by a runner carries no trace of the runner at all. The `speaker_*` fields are comment-only, unverified free strings, and not a joinable identity.

This doc makes the **actor** formal as a pair:

```
actor = ( account → users,  sub_identity → sub_identities | NULL )
```

- **Account** = the responsible human; every existing `actor` / `created_by` / `updated_by` column keeps this meaning, and all permission checks keep reading it.
- **Sub-identity** = a non-human principal under exactly one account: a runner today, with `mcp` / `cloud_agent` / `assistant` kinds for the other agent surfaces.

## Key decisions

- One new `sub_identities` table (soft-delete only, workspace-scoped, 1:1 to `runner` with `SET_NULL` since runner rows are hard-deleted) — **no `actors` table**, no bot `User` rows.
- One nullable sibling column beside each existing account column (`*_sub_identity` on `UserAuditModel`, `actor_sub_identity` on comments/activities/reactions/votes). No backfill: NULL = the human acted directly.
- Resolved once per request from the `X-Pi-Dash-Run-Id` header the CLI **already sends on every write**, reusing `run_belongs_to()`; stamped centrally in `BaseModel.save()` next to `created_by`. No CLI / daemon / prompt changes.
- Server-side agents use an `acting_as(sub_identity)` context manager beside `crum.impersonate`; Celery tasks take a `sub_identity_id` kwarg.
- API is additive (`actor_info` with both halves); `speaker_*` become server-filled / read-only.
- Rollout in four independently shippable phases; old `pidash` binaries degrade to today's behaviour.

## Test plan

- [ ] Review the schema (§3.2–3.3) and resolver rules (§3.4)
- [ ] Confirm the open questions in §8 (independent rename, admin visibility, shared runners)


irichard00:
Design review of `.ai_design/create_sub_identity/design.md` (PR #431, `docs/sub-identity-design`). I checked the doc against the code on `main` (`ea223760`).

**Verdict: the direction is right. Please revise before implementation.** The (account, sub-identity) pair, having no `actors` table, rejecting a bot `User` per runner, and resolving from `X-Pi-Dash-Run-Id` are all good calls. Most of the doc's factual claims hold up. The CLI sends the header on every mutating request (`runner/src/api_client.rs:257`). `run_belongs_to` and `resolve_moved_by_run` are as described (`api/views/issue.py:941`, `:957`). The workpad and yield endpoints are `BaseAPIView`s, so hooking `BaseAPIView.initial()` covers every CLI write. `app/serializers/issue.py:960` is `fields = "__all__"`, and `speaker_*` is not in `read_only_fields`. Runner rows are hard-deleted by `runner_delete.py`.

Six things should change in the doc first. The first two are the most important.

**1. Someone other than the runner owner can already trigger a run on that runner. §5 treats this as a future shared-runners problem.** A `PRIVATE` runner serves runs whose creator, issue creator or assignee is its owner (`runner/services/matcher.py:375` `pod_has_runner_for_issue_principal`). `run_belongs_to` explicitly accepts `run.created_by` ("Run AI / Comment & Run, the person who asked"). Example: Alice presses Run AI on an issue assigned to Rich, and Rich's runner picks it up. The CLI authenticates with Rich's machine token, so the account is Rich. The UI would then say "rich-mbp-1 · on behalf of Rich M", but Alice asked for the work.

The invariant `sub_identity.owner == account` still holds, because the token and the runner have the same owner. What's wrong is the claim that "on behalf of" names the human who wanted the work. The doc needs to choose one of these:
- (a) "On behalf of" means the runner owner, who is accountable. The asker is shown separately, e.g. "requested by Alice" via `speaker_agent_run_id → run.created_by`.
- (b) The account becomes `run.created_by`. That breaks the token-owner model.

I'd recommend (a), and rewording the last paragraph of §5. It is not a deferred problem.

**2. Adding the columns to every table is too broad, and the migration isn't as cheap as the doc says.** About 116 models inherit `AuditModel → UserAuditModel`, so the mixin change adds about 232 FK columns. Only issues, comments, activities, reactions and votes are ever read (§3.6/§3.7). Two costs follow:
- *The migration.* Django adds the column and then a separate `FOREIGN KEY` constraint. Validating that constraint scans the whole table under a lock, even when every value is NULL. So the change is metadata-only for the column but not for the constraint, and it happens ~232 times.
- *Deletes.* `SET_NULL` is enforced by Django's delete collector, not by Postgres. Deleting a sub-identity, which happens when a user is deleted, would run about 232 `UPDATE … WHERE created_by_sub_identity_id IN (…)` statements against unindexed columns.

Suggestion: add explicit columns only on the tables the design actually reads (`Issue` plus the explicit-actor tables). Have `BaseModel.save()` stamp only when the model has the field. If you want to future-proof the other tables, use `db_constraint=False` + `on_delete=DO_NOTHING` there. Sub-identities are soft-delete-only anyway.

**3. The activity task has 24 caller files, and none of them will pass the new kwarg by default.** `issue_activity.delay(...)` is called from 24 non-test files. The task builds its rows with `IssueActivity(...)` ×44 and then `bulk_create`, so the stamp in `save()` never runs. The optional `sub_identity_id=None` kwarg is right for in-flight tasks. The risk is that it fails silently: any call site that isn't updated records the human, and no test catches it. Suggestion: add a thin enqueue helper (or override `apply_async` on the task) that fills in `get_current_sub_identity()` automatically. That way, stamping by default doesn't depend on updating 24 files. The same applies to `model_activity` and webhooks.

**4. The resolver shouldn't turn every write into a possible 400.** Today only issue patch refuses a foreign run id. Step 3 of the resolver would extend that 400 to every `/api/v1` write, including comment, workpad and yield. Attribution is best-effort: step 4 already drops a mismatched sub-identity to `None`. Suggestion: outside `resolve_moved_by_run`, a foreign or unknown run should resolve to `None` and log a warning, not reject the write. Also, `crum` already stores the request, so `get_current_sub_identity()` can read `request.sub_identity` through `crum.get_current_request()` plus the `acting_as` stack. That removes a second thread-local and the `finalize_response` cleanup it needs.

**5. Details of the stamp in `BaseModel.save()` that the doc should spell out:**
- On update, always write `updated_by_sub_identity`, including clearing it back to NULL when a human edits a row an agent last touched.
- `save(update_fields=[...])` is widespread. Like `updated_by`, the sub-identity column is silently skipped unless it is added to `update_fields`. Decide whether you're matching that behaviour or fixing it.
- Say what happens on the explicit `created_by_id=` path and the `disable_auto_set_user` path.

**6. The runner lifecycle hooks miss some paths.**
- *Renames.* `runner.name` changes in two places: the PATCH at `runner/views/runners.py:379`, and re-enrollment at `runner/views/enrollment.py:346`. It also only gets created in one place, `enrollment.py:726`. I couldn't find a separate machine-control create path; if §3.2 means `enrollment.py:726`, please name it. Instead of mirroring `display_name` on every write, read `runner.name` live while `runner_id` is set and snapshot it only on revoke. Renames then need no hook.
- *Revocation.* `Runner.revoke()` sets `revoked_at` without deleting the row. Runners are also deleted by the `owner`/`workspace` CASCADE, which doesn't go through `runner_delete.py`. Either treat `kind=runner AND runner_id IS NULL` as revoked, or use a `post_delete` signal, so "(removed)" still renders.

**Smaller notes**
- *Public boards.* `apps/space` would show runner names on public issues, and runner names are often hostnames (`rich-mbp-1`). Consider a generic "Agent" label there, or add this as open question 4.
- *`CONCURRENTLY` indexes* need `atomic = False` and their own migration. Worth a line in §6.
- *§3.8* maps `speaker_type` (`human/agent/system/integration`) to `kind` (`runner/mcp/cloud_agent/assistant`). Please give the actual mapping table.
- *§7* should also test: a human edit clears `updated_by_sub_identity`; a runner deleted through a CASCADE still renders as removed; a Run AI run by a non-owner (point 1).

**The doc's open questions:** I agree with no independent rename (Q1) and with admins seeing only what acted in projects they can see (Q2). On Q3: part of it isn't deferred. Runs by a non-owner exist today (point 1).

PR #432, opened 2026-09-21
feat(cli): default auth login to the hosted cloud instead of prompting

irichard00:
### Description

The released install one-liner runs straight into `pidash auth login`, which
stopped on a free-text `Cloud URL:` prompt. Two problems with that:

- Every user of AI Republic-hosted Pi Dash had to type our own production
  hostname (`https://pidash.airepublic.com`) in by hand.
- Headless installs got no prompt at all, just
  `no cloud URL configured — pass --url …`.

`resolve_cloud_url` now falls back to a new crate constant `DEFAULT_CLOUD_URL`
(`runner/src/lib.rs`) when there's no `--url` and no `[daemon].cloud_url`,
replacing both the prompt and the headless bail. Resolution order is otherwise
unchanged:

1. `--url` flag — still wins over everything
2. this host's existing `[daemon].cloud_url` — an enrolled self-hosted box
   stays on its own cloud across re-logins
3. `DEFAULT_CLOUD_URL` ← previously the prompt

The resolver also returns a `CloudUrlSource` so the flow can say where the URL
came from, and prints the destination **only** when it chose for the user:

```
Connecting to Pi Dash at https://pidash.airepublic.com
(Self-hosted? Cancel and run `pidash auth login --url <YOUR-URL>` instead.)
```

`write_cli_token`'s cross-cloud guard is untouched, so this can't silently
rebind an already-enrolled host.

Side effects, both intended:

- **Headless now works unattended** — `pidash auth login --no-browser` needs no
  flags. `install.sh`, `install.ps1`, and the quickstart drop `--url <URL>`
  from their headless instructions.
- **A self-hoster who runs the one-liner and hits Enter reaches prod.** That's
  the trade for removing the prompt. The printed notice catches it, and a
  re-run with `--url` fixes it cleanly since no config has been written yet.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] Feature (non-breaking change which adds functionality)
- [x] Improvement (change that would cause existing functionality to not work as expected)
- [ ] Code refactoring
- [ ] Performance improvements
- [x] Documentation update

### Test Scenarios

Automated — 5 new unit tests in `runner/src/cli/auth/login.rs`:

- [x] `defaults_to_hosted_cloud_when_nothing_is_configured` — the install
      one-liner's path: no flag, no `config.toml` → `DEFAULT_CLOUD_URL`
- [x] `default_cloud_url_passes_transport_validation` — the default clears the
      same https-only bar as a user-supplied URL
- [x] `explicit_url_flag_wins_and_trailing_slash_is_trimmed`
- [x] `existing_config_url_is_reused_over_the_default` — re-login on an
      enrolled self-hosted host stays on that cloud
- [x] `explicit_url_flag_wins_over_existing_config`
- [x] Full suite green: 497 lib tests + all integration tests, 0 failures
- [x] `cargo clippy --all-targets` clean; `cargo fmt --check` shows no diff in
      the touched files (repo has pre-existing drift elsewhere, left alone)
- [x] `sh -n runner/install.sh` passes

Manual, before merge:

- [ ] Fresh host, no `~/.config/pidash/config.toml`: run the release
      `install.sh` and confirm it goes straight to the device-code block
      against prod, with no URL question
- [ ] Same, headless (`ssh` without `-t` / Docker without `-t`): confirm
      `pidash auth login --no-browser` prints an approvable URL instead of
      erroring
- [ ] Self-hosted: `pidash auth login --url https://<instance>` on a clean host
      still enrolls against that instance
- [ ] Re-login on a host already enrolled to a self-hosted instance stays on
      that cloud and prints no "Connecting to…" line
- [ ] Windows: `install.ps1` one-liner reaches the same flow

### References

Follow-up not in this PR: the TUI register form still defaults its cloud-URL
field to `http://localhost` (`runner/src/tui/views/general.rs:68`).


irichard00:
Review pass (Pi Dash PDASHOSS01-194). No blocking findings.

I checked:
- `resolve_cloud_url` order is still `--url` > `[daemon].cloud_url` > `DEFAULT_CLOUD_URL`. No other code called the removed prompt or relied on the old headless bail.
- `write_cli_token`'s cross-cloud guard hasn't changed, so the default can't quietly rebind an enrolled host.
- Locally on the branch head, the full `cargo test` passes (492 lib tests plus all integration tests), `cargo clippy --all-targets` is clean, and `sh -n install.sh` passes. CI `cargo test + clippy` is green.

One fix I pushed (dfc0fee7): the quickstart told self-hosters to run `--url` "afterwards". Once the device code is approved against the hosted cloud, config.toml is bound to it and the cross-cloud guard refuses a later `--url` login. The text now says to cancel with Ctrl-C before approving.

Not in scope: the TUI register form still defaults to `http://localhost` (`runner/src/tui/views/general.rs`), as the PR notes. The manual test-plan items (fresh host, headless, Windows) are left for the test stage.

PR #433, opened 2026-09-21
PDASHOSS01-196 Prompt: surface blocked_by / blocking (and other directional relations) in the "Work item relationships" section

irichard00:
Closes PDASHOSS01-196 (child of PDASHOSS01-195, "inform and wake, no hard gate").

## Problem
The agent prompt's **Work item relationships** section showed only the parent chain, direct children, and `relates_to` items. `blocked_by`, `blocking`, `start_*`, `finish_*`, and `implemented_by`/`implements` rows in `IssueRelation` never reached the agent. There's no dispatch gate, so the agent's own judgement is the only thing that stops it from starting work while a blocker is still open.

## Change
- **`prompting/context.py`**: new `_directional_relations_context` collector. It runs one query over both link directions, and the other end's view comes from `IssueRelationChoices._REVERSE_MAPPING`. It dedupes per type, skips self and soft-deleted relations and targets, and caps each type at `_MAX_RELATIONSHIP_ITEMS`. Each item is `{identifier, title, state, state_group}`. `build_context` exposes `blocked_by`, `blocking`, `other_relations` (each item has a `relation` label), `open_blockers`, and `has_open_blockers`. A blocker counts as open when its state group is not `completed` or `cancelled`. The query stays local because `open_blockers()` from PDASHOSS01-197 hasn't landed yet, so 197 can refactor onto it.
- **`sections/relationships.md`** (plus the matching legacy fragment `02_relationships.md`) gains three groups:
  - **Blocked by (must be done first)**, with a one-line warning naming the open blockers.
  - **Blocking (waiting on this item)**.
  - **Other relations**.

  The required-reading directive now covers these groups. When there are open blockers it also carries decision guidance, not a rule: the agent either proceeds and records its assumption, or writes `Waiting on: <IDs>` in the workpad, posts one short comment, and exits as a noop. The section is still omitted when the issue has no relations at all.
- **Migration `0007`** force-reseeds the default PromptTemplate, the same approach as `0006`.
- **`validation.py`** sample contexts include the new keys, so override validation exercises them under `StrictUndefined`.
- No recipe changes.

## Tests
- `tests/unit/prompting/test_context.py`: both link directions, reverse mapping for all 6 other types, `relates_to`/`duplicate` excluded, one entry per pair, self skipped, cap, soft-deleted relation and target skipped, empty case, `has_open_blockers` true and false.
- `tests/unit/prompting/test_composer.py`: an open blocker renders the group, the warning, and the guidance; closed blockers get no warning; groups are omitted when empty; `blocking` and other relations render; the warning pluralises.
- `tests/contract/prompting/test_compiled_preview.py`: the preview endpoint shows Blocked by / Blocking for a real issue.

`pytest pi_dash/tests/unit/prompting/ pi_dash/tests/contract/prompting/test_compiled_preview.py` → 164 passed. `ruff check` is clean on the touched paths.

## Note (out of scope)
`utils/issue_relation_mapper.get_actual_relation` stores a user-created `implements` link as `implemented_by` without swapping the endpoints, unlike `blocking`/`start_after`/`finish_after`. So an `implements` link made from the UI reads back reversed, both here and anywhere else that reads the stored type. This existed before this PR and isn't changed by it.


irichard00:
**Automated review (PDASHOSS01-196): approved, no blocking findings.**

Checked:
- `_directional_relations_context`: one query over both endpoints; forward types come from `_REVERSE_MAPPING` keys, so the reverse lookup is total. Self links, soft-deleted relations (default manager) and soft-deleted issues on either end are filtered out. Each type is capped at `_MAX_RELATIONSHIP_ITEMS`. `open_blockers` means state group not in {completed, cancelled}.
- Section and fragment bodies are identical. The section still renders nothing when there are no relations of any kind. The Blocked-by / Blocking / Other groups and the guidance each appear only when their data is present.
- The wait exit (`Waiting on: <IDs>` + `pidash run yield --outcome waiting_on_external`) matches what the PDASHOSS01-198 branch parses and treats as a waiting run (`WAITING_RUN_STATUSES`).
- Migration `prompting.0007` depends on `0006`, and no other open pdashoss01 branch adds a `0007`. `validation.py` samples cover the new keys. The preview contract test covers the endpoint.
- CI (pytest, Lint API) is green on 7edb59b2 and the branch merges cleanly into main.

Non-blocking: the `implements` read-back inversion from `get_actual_relation` predates this PR and is already called out in the description as a follow-up.

PR #435, opened 2026-09-21
PDASHOSS01-198 Orchestration: wake dependents when a blocker completes, and pause cadence ticks while an agent has chosen to wait

irichard00:
Part of PDASHOSS01-195. This change makes waiting on a blocker cheap. The agent still decides whether to wait, and the platform honours that decision and wakes the agent as soon as a blocker changes.

> **Stacked on #434 (PDASHOSS01-197).** It uses `orchestration/blockers.py` from that PR. Until #434 merges, this diff also includes that commit (`87b6242c`). Review only the top commit, `PDASHOSS01-198 …`.

## What changes

**A. Wake on blocker completion.** The `Issue` post-save hook in `orchestration/signals.py` now watches for an issue moving *into* a `completed` / `cancelled` group. When that happens, it enqueues `orchestration.wake.wake_dependents` on commit.
- For each `dependents(blocker)` in a ticking state with an armed clock, the task runs `fire_tick(ticker_id, trigger="blocker_completed")` now instead of waiting for the cadence.
- It uses the new `AgentRunTrigger.BLOCKER_COMPLETED`. The wake counts against the budget like a tick. It runs as the system bot and is admitted as automatic.
- It skips if a run is already active, and it defers to a queued entry run.
- Backlog / Todo dependents are never moved and never run.
- Re-saving a blocker that is already closed, or moving it Done → Cancelled, does not fire again.

**B. Pause cadence while waiting.** `fire_tick` has a new pre-claim skip, `reason=waiting-on-blockers`. The skip leaves `used` untouched and pushes `next_run_at` out one interval, so the scanner doesn't retry every minute. All of these must hold for the skip:
- the last run finished and yielded `waiting_on_external` (or legacy `noop`). This is the yield that #433's guidance asks for;
- the workpad has a `Waiting on:` marker (`parse_waiting_on`). The parser accepts bullets, bold, and any case, and ignores fenced code;
- every listed ID resolves to a **still-open `blocked_by` target** of the issue. An unknown ID, an unrelated issue, or one that is already closed does not pause, because nothing would ever wake it;
- no human comment has arrived since that run ended. Agent-spoken and bot comments don't count;
- less than `Project.agent_wait_max_pause_seconds` has passed since that run ended. The default is 7 days, and `0` turns the pause off.

Any newer run ends the pause, because the waiting run is then no longer the latest. That covers a human move, Comment & Run, Run AI, and a wake. A queued entry run is never paused.

**C. Visibility.** A woken dependent gets an `agent_wake` activity entry, recorded as `Woken: WEB-1 completed`. The web activity feed renders it as *"Pi Dash woke the agent: blocker WEB-1 was completed."* Skip, resume (`max-pause-elapsed`), and wake decisions are logged.

**Prompt.** The workpad template gains `- Waiting on: <none, or …>` plus a note on how Pi Dash reads it. The session framing has a new "why this run started" branch for `blocker_completed` runs.

## Migrations
- `db/0165_project_agent_wait_max_pause`: new column with default 604800.
- `runner/0027_agentrun_trigger_blocker_completed`: choices only. I wrote it by hand because `makemigrations runner` also picks up drift that is already on main.
- No prompting reseed. `sections/*.md` are read from disk, so this also avoids a clash with #433's `prompting/0007`.

## Tests
`pi_dash/tests/unit/orchestration/test_wake.py`: 43 tests covering the parser, each pause condition, the fire_tick skip and the forced wake, and wake_dependents. It also has an integration test that goes through signal → task → `fire_tick`: B is blocked by A and C, A → Done wakes B, and C → Done wakes B again.

- `unit/` full suite: 1467 passed. 3 failures, all environment-only: the `fastmcp` module is missing in the container, plus the known `test_valid_agents_matches_runner_schema` mount artifact.
- move / workpad / runner / loop contract tests: 63 passed.
- `contract/prompting`: passed.


irichard00:
**Review of PR #435 (PDASHOSS01-198), top commit `fbe09880`. `87b6242c` belongs to #434 and I didn't review it here.**

Approved. I found no defects that block merging. CI is green on `fbe09880`, pytest included.

What I checked:
- **`fire_tick` forced mode.** A forced tick skips only the `next_run_at` check and the waiting pause. The cap / disarm, ticking-state, active-run and no-prior-run guards all still apply. A pending entry wins over a wake and serves it. The claim / rollback path is unchanged, and a wake counts against the budget.
- **The pause.** It runs only for a non-forced tick with no pending entry. It re-times `next_run_at` by one interval and doesn't touch `used`. `waiting_pause` requires all of these: the latest run is finished with status `waiting_on_external` / `noop`; the marker is present; every listed ID resolves to a live, still-open `blocked_by` target; there's no human comment since the run ended; the max pause hasn't elapsed (`0` turns it off).
- **Signal.** The wake is enqueued only on a transition from open into closed, via `on_commit`. Broker errors are swallowed and counted. The task re-checks that the blocker is still closed before waking anything.
- **Trigger plumbing.** `AUTOMATIC_ISSUE_TRIGGERS` and `_MACHINE_TRIGGERS` cover every earlier `== TICK` comparison: creator resolution and cloud admission. The runner migration is choices-only.
- **Activity.** The issue activity endpoint doesn't filter out `agent_wake`, so the new web renderer gets the entry.

Non-blocking notes:
1. If a blocker closes while the dependent has a run in flight, that wake is dropped because the active-run-exists skip applies, as the spec asks. The issue doesn't stay stuck, though. The pause needs every listed blocker to still be open, so the dependent's next cadence tick (at most 3 h later) fires normally.
2. State changes made through a queryset `.update()` bypass `post_save`, so they won't wake dependents. The existing transition handler has the same limitation, so this isn't new to this PR.

